PR #29, opened 2020-06-19
koji: improve log messages on missing RPM(s)

rohanpm:
When we can't find a requested RPM on the filesystem, previously
we'd log that RPMs were not found but it wasn't clear what it
means. Adjust the error message to include the paths of the missing
RPM(s) to make it more obvious.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- n/a
+### Changed
+- Improved log message details in some cases of error
 
 ## 1.0.0 - 2020-06-16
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- n/a
+### Changed
+- Improved log message details in some cases of error
```
chandwanitulsi, line 11: Minor: How about adding "koji error" here?

rohanpm: I was a little vague here because there are other pull requests also about improving logs in other cases of error - so I just wanted one note covering them all.  Unfortunately this setup having just one CHANGELOG file is quite annoying here, as separate PRs adding text at the same place will likely conflict, so there's a disincentive to write a separate point for every PR.

PR #32, opened 2020-06-24
Introduce PushItem.with_checksums for calculating sums

rohanpm:
As explained in doc string. Calculating checksums is potentially
expensive, and so it is not filled in implicitly by all pushsource
backends. The caller is able to control when sums are calculated

This utility method makes it simple to obtain checksums only when
needed, e.g. a Pub-style multithreaded "computing checksums" step
could be easily implemented via a threadpool executor by:

    items = executor.map(lambda i: i.with_checksums(), items)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ def get_requirements():
 
 setup(
     name="pushsource",
-    version="1.1.0",
+    version="1.2.0",
     packages=find_packages("src"),
     package_dir={"": "src"},
     include_package_data=True,
--- a/src/pushsource/_impl/model/base.py
+++ b/src/pushsource/_impl/model/base.py
@@ -1,7 +1,15 @@
+import hashlib
+import os
+import logging
+
 from frozenlist2 import frozenlist
 
 from .. import compat_attr as attr
 from .conv import md5str, sha256str, upper_if_str, instance_of_str, optional_str
+
+
+LOG = logging.getLogger("pushsource")
+CHUNKSIZE = int(os.environ.get("PUSHSOURCE_CHUNKSIZE") or 1024 * 1024 * 16)
 
 
 @attr.s()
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -21,7 +21,7 @@ def get_requirements():
 
 setup(
     name="pushsource",
-    version="1.1.0",
+    version="1.2.0",
```
querti, line 24: Should this be here? If yes then shouldn't it also be in the changelog?

src/pushsource/_impl/model/base.py
```diff
@@ -78,3 +96,67 @@ class PushItem(object):
...
+        sha256_hasher = hashlib.sha256()
+
+        LOG.debug("Start read: %s", self.src)
+
+        with open(self.src, "rb") as src_file:
+            while True:
+                chunk = src_file.read(CHUNKSIZE)
+                if not chunk:
+                    break
+                md5_hasher.update(chunk)
```
querti, line 155: Probably dumb question, but is it possible that only one of these checksums will already be present? If it could, then some computation can be saved.

setup.py
rohanpm: I did intentionally put this here, but there isn't really a well-established precedent for how to handle it.

These projects are meant to follow SemVer, so adding an API (like I've done here) means that the next version number should have the middle number bumped.  So: that bump can be included in the PR which adds the new API, which makes it the responsibility of the developer making a change to evaluate what it means with respect to SemVer.  Or, the bump can be *not* included in the PR, in which case whoever does the next release needs to inspect all the changes and figure out which of the three version numbers needs to be bumped per SemVer rules..

I think it's better to have the bump included in the PR because to me it seems like there's a high chance of release masters making a mistake when deciding the next version or even just forgetting to think about SemVer altogether.  I would like the whole process to be improved somehow though.

About whether it should be in the changelog, no I don't think so... I mean, there are arguments for and against, but 1.2.0 is still unreleased so it does make sense to me that the changelog still just says "Unreleased".  Note the 1.2.0 isn't set in stone... for example someone else could make a PR with backwards-incompatible changes meaning the next version should actually be 2.0.0... and it's probably more confusing to have a version number appear in a changelog, then disappear later, than it is to just group all unreleased content under "Unreleased".

src/pushsource/_impl/model/base.py
rohanpm: Yeah, it is possible.  I've assumed (without testing) that the time reading the file here far outweighs the time calculating the checksum, meaning it's not worth trying to skip individual checksum types if any type needs to be calculated.  I'll test that assumption and report back here.

rohanpm: I tested it and I was wrong, based on https://gist.github.com/rohanpm/2c1662e39fb138869cd4901cc9b2f9d4, there could be a substantial gain from avoiding the recalculation of checksum types we already have.  So this should be revised.

PR #33, opened 2020-06-30
tests: pin all dependency versions and enable dependabot

rohanpm:
List specific versions of every dependency to be used in CI
in the test requirements files. This ensures that CI cannot break
due to backwards-incompatible changes in dependencies.

To ensure that these are kept up-to-date over time, dependabot
is enabled. It will create PRs automatically when new versions of
dependencies are available.


rohanpm:
This is a bit of an experiment.

We've used dependabot successfully in exodus-lambda project for a while, you can see some example PRs at https://github.com/release-engineering/exodus-lambda/pulls?q=is%3Apr+is%3Aclosed+label%3Adependencies.  But that project is easier in some respects, in particular it doesn't have to support multiple Python versions.  There doesn't seem to be a way to tell dependabot that we want to test py2 with some older library versions, and py3 with newer versions, so I'm not sure how well this is going to work out in practice.  It might need tweaking later.

There's an (internal) doc with more notes about what's going on here: https://docs.engineering.redhat.com/x/VCulC
If it works out OK for this project then I will share it more broadly and propose it for other projects later.

.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+- package-ecosystem: pip
+  directory: /
+  schedule:
+    interval: daily
```
yashvardhannanavati, line 6: Can I ask if there's a reason for this frequency? It can potentially be very noisy, no?

yashvardhannanavati: On second thoughts, I think it should be fine.

yashvardhannanavati:
👍

.github/dependabot.yml
rohanpm: My idea here is to start with more frequent so that we get more experience with this sooner.  I would not be surprised if we later settle on weekly as the best schedule, but seeing as this is a little experimental, it's not ideal if it only does something once a week right now.

PR #57, opened 2020-07-27
ci: make dependabot skip legacy constraints

rohanpm:
We want to have CI covering both:

- latest released version of all deps, and:

- oldest supported versions of selected deps (e.g. for the oldest
  RHEL version where we expect the code to be used) - we use
  a constraints file to achieve this.

Problem: dependabot doesn't understand the setup, and it tries to
update constraints-legacy.txt as if we want the latest version of
all deps there as well, breaking the purpose of this file.

Apparently, there is no supported option to make dependabot skip
the updating of the constraints file, see:
https://github.com/dependabot/dependabot-core/issues/1657

However, from reviewing the code I can see that dependabot's python
support will only find requirements/constraints files ending in
".txt" or ".in". The underlying tools don't actually require this,
so let's name the constraints file as something else to remove it
from dependabot's purview.


rohanpm:
This is an example of the kind of PR we don't want dependabot to make: https://github.com/release-engineering/pushsource/pull/49

This change should prevent them in future.

rohanpm:
@chandwanitulsi , @rbikar can you have a look when you have a moment?

PR #102, opened 2020-11-01
cleanup: drop useless channel dumps code

rohanpm:
Pushing channel dumps has not been done since 2019-03, and is not ever
intended to be used again. Hence, this code was obsolete before it was
written.

Rather than carrying support for this kind of push item which is not
ever expected to be used, let's clean it up before integrating "staged"
source into Pub.


rohanpm:
This is a draft at the moment as I haven't got confirmation that it's obsolete.

rohanpm:
Confirmed in SPMM-4095 that it's OK to do this.

midnightercz (approved):

rbikar (approved):

PR #105, opened 2020-11-02
Improved logging when schema validation fails

rohanpm:
Previously, wherever we used jsonschema, if any validation failed
we would simply allow the ValidationError to be propagated.

This works, but is not very user-friendly. The problem is that
the message on a ValidationError often will not make sense to the
end-user. It is a developer-oriented message and it often
requires understanding json schema in order to know which piece
of data failed validation.

Let's improve this a bit. We will still raise the same exceptions,
but before we do that, let's log a more user-friendly message which
can hopefully allow the user to at least figure out which object
failed validation. We'll do this via a new Validator class which
looks at the exceptions generated by jsonschema.

This will result in user-oriented messages logged when validation
fails along the lines of:

  my-repo/FILES/foobar.iso (in pub-mapfile.json) is not valid

...so the user knows at least which piece of metadata was invalid,
while the full details of the validation error are still available
in the raised exception.


rohanpm:
Prepared this while testing this library with Pub, as I noticed that the logging behavior in some error cases was less user friendly than before.

nathanegillett (approved):

rbikar (approved):

PR #108, opened 2020-11-04
erratum: rename "from_" to "from" (with alias available)

rohanpm:
Erratum objects have a "from" field, which unfortunately clashes
with a Python keyword.

It's not possible by default to use an attribute name which
matches a Python keyword. As a workaround for this, the attribute
was previously named simply "from_".

Although that's a simple solution, while writing code using this
library I've come to believe it was a mistake.

The problem is that these push item classes are mostly used for
serialization and deserialization of objects coming from various
sources, all of which are going to be using the preferred attribute
name, "from".

This implies that, with the old solution, almost every usage of
ErratumPushItem class would have to be accompanied with some
"from" <=> "from_" renaming hack. In the worst case, this renaming
could be forgotten and the "from_" name would leak into other tools
and services due to this library.

Hence, although it requires a bit of tricky code extending attrs
to get it working, I believe it's worthwhile to complicate this
library a bit so that the attribute can be genuinely named "from".
This means the complexity due to the keyword clash can be kept
within this library rather than spilling into every user of the
ErratumPushItem class.

While constructing or reading instances of the class, the "from_"
name may still be used as an alias, so this is a
backwards-compatible change.


rohanpm:
Due to "weirdness" of this change, I want to keep it a draft until I have a chance to actually try it with Pub.

rohanpm:
Have experimented with this in Pub, seems to be working fine.

chandwanitulsi (approved):

PR #109, opened 2020-11-08
staged: accept metadata for legacy unsupported content

rohanpm:
The intended behavior for old unsupported content types such as
DOCKER is to log a warning and ignore discovered push items.

However, it is likely that the metadata file for the push also
has entries for those files. In that case, these must be able to
pass validation, otherwise we'll get an error rather than a
warning.

To make that work, add a loose schema for those unsupported types.
It does not validate any attributes on the metadata, as the goal
is simply to ignore these entries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- n/a
+### Changed
+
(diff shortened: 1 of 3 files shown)


midnightercz (approved):

rohanpm:
@rbikar or @chandwanitulsi could you please review, as code owners?

nathanegillett (approved):

chandwanitulsi (approved):

PR #114, opened 2021-02-19
Add billing codes to staged AMI push item

rbikar:
This commit adds billing codes field for AMI staged push item schema
and model.

This field is optional and it's a dict consisting of
- "name", that is arbitrary string used for compiling unique AMI image name 
- "codes" list of actual billing codes that should be for AMI registering.

### DIFF
--- a/src/pushsource/_impl/backend/staged/staged_ami.py
+++ b/src/pushsource/_impl/backend/staged/staged_ami.py
@@ -1,7 +1,7 @@
 import os
 import logging
 
-from ...model import AmiRelease, AmiPushItem
+from ...model import AmiRelease, AmiPushItem, AmiBillingCode
 from .staged_base import StagedBaseMixin, handles_type
 
 LOG = logging.getLogger("pushsource")
@@ -33,6 +33,17 @@ class StagedAmiMixin(StagedBaseMixin):
 
         release = AmiRelease(**release_kwargs)
 
+        billing_code_md = attributes.get("billing_code")
+
+        billing_code_attrs = ["name", "code"]
+        billing_code_kwargs = {}
+
+        for key in billing_code_attrs:
+            if key in billing_code_md:
+                billing_code_kwargs[key] = billing_code_md[key]
+
+        billing_code = AmiBillingCode(**billing_code_kwargs)
+
         image_attrs = [
             "type",
             "region",
@@ -43,7 +54,7 @@ class StagedAmiMixin(StagedBaseMixin):
             "sriov_net_support",
             "ena_support",
         ]
-        image_kwargs = {"release": release}
+        image_kwargs = {"release": release, "billing_code": billing_code}
         for key in image_attrs:
             if key in attributes:
                 image_kwargs[key] = attributes[key]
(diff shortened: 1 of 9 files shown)


src/pushsource/_impl/backend/staged/staged_ami.py
```diff
@@ -33,6 +33,17 @@ def __push_item(self, leafdir, metadata, entry):
 
         release = AmiRelease(**release_kwargs)
 
+        billing_code_md = attributes.get("billing_code")
+
+        billing_code_attrs = ["name", "code"]
+        billing_code_kwargs = {}
+
+        for key in billing_code_attrs:
+            if key in billing_code_md:
```
rohanpm, line 42: Am I missing something or is it going to crash if "billing_code" was absent in the metadata file?

It will usually be absent, so then billing_code_md will be None, and 'key in billing_code_md' would crash.

If I'm right: it would have been better to update the test to cover both cases, rather than updating it to cover only the new case of billing_code being present.

rohanpm (changes requested):

src/pushsource/_impl/backend/staged/staged_ami.py
rbikar: This is currently on hold, the specification of required functionality was not properly described. Not sure if it should be optional or required.
But you're now it's gonna crash with missing billing code.

rbikar:
Requirement for AMI billing codes were clarified so I made this field optional and it contains list of billing codes instead of one code.

src/pushsource/_impl/model/ami.py
```diff
@@ -41,6 +41,17 @@ class AmiRelease(object):
     """Release type, typically "ga" or "beta"."""
 
 
+@attr.s()
+class AmiBillingCodes(object):
+    """Billing codes associated with an AMI."""
+
+    name = attr.ib(type=str, default=None, validator=instance_of_str)
+    """Billing codes name, for example: Hourly2, arbitrary string for making image name unique"""
```
rohanpm, line 49: Minor: the existing doc string style is to end the doc string with a ".", making it more of a full sentence rather than sentence fragment.  It'd be best to stick with that.

src/pushsource/_impl/model/ami.py
```diff
@@ -41,6 +41,17 @@ class AmiRelease(object):
...
+
+    codes = attr.ib(type=list, default=None, validator=optional(instance_of(list)))
```
rohanpm, line 51: Can you please use frozenlist for this to ensure the object remains immutable?

There are examples in other classes, like "content_types" in ErratumPushItem.  Basically you want to set 'default' and 'converter'.

src/pushsource/_impl/schema/staged-schema.yaml
```diff
@@ -212,6 +216,20 @@ definitions:
...
+    ami_billing_codes:
+        # Billing codes for an AMI
+        type:
+        - object
+        - "null"
+        properties:
+            name:
+                type: string
+            codes:
+                type: array
```
rohanpm, line 228: I think this will allow an array of any type of object, can we please make it an array of string?

rohanpm (changes requested):
Can it please be mentioned in CHANGELOG.md as well?

chandwanitulsi (approved):

rohanpm (approved):

PR #125, opened 2021-06-01
refactor: move modulemd filtering from errata source to koji source

rohanpm:
ET requests modules by filename, e.g. modulemd.x86_64.txt.

Previously, we had errata source query koji source and then filter
the returned values according to ET's desired filenames. Let's
instead make koji source natively support filtering by way of a new
parameter.

Motivation: koji source will soon need to start parsing these
modulemd files so that the NSVCA can be put onto push items.
If we do this for files which will be then filtered by the errata
source, we're negatively impacting the performance and reliability
by loading data we don't need. It's better if we can tell koji
source in advance exactly which files we're interested in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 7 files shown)


rohanpm:
I'm going to reopen this using a source branch within this repo, so I can set up stacked pull requests.

PR #126, opened 2021-06-03
refactor: move modulemd filtering from errata source to koji source

rohanpm:
ET requests modules by filename, e.g. modulemd.x86_64.txt.

Previously, we had errata source query koji source and then filter
the returned values according to ET's desired filenames. Let's
instead make koji source natively support filtering by way of a new
parameter.

Motivation: koji source will soon need to start parsing these
modulemd files so that the NSVCA can be put onto push items.
If we do this for files which will be then filtered by the errata
source, we're negatively impacting the performance and reliability
by loading data we don't need. It's better if we can tell koji
source in advance exactly which files we're interested in.

### DIFF
--- a/src/pushsource/_impl/backend/koji_source.py
+++ b/src/pushsource/_impl/backend/koji_source.py
@@ -102,6 +102,7 @@ class KojiSource(Source):
         dest=None,
         rpm=None,
         module_build=None,
+        module_filter_filename=None,
         signing_key=None,
         basedir=None,
         threads=4,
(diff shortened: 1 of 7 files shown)


rohanpm:
Because I've had a lot of pull requests for this repo with some dependencies between them, I'm trying a "stacked pull request" workflow at the moment.  This is the reason why this PR is using this repo as the source, rather than using my fork.

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -244,6 +265,15 @@ def _push_items_from_rpm_meta(self, rpm, meta):
             )
         ]
 
+    def _module_filtered(self, file_path):
+        ok_names = self._module_filter_filename
+        filename = os.path.basename(file_path)
+        if ok_names != None and filename not in ok_names:
```
rbikar, line 271: Not a big deal I guess, but `ok_names is not None ` should be used when comparing to None or probably only ` if ok_names and filename not in ok_names:` can be used as well.

rbikar (changes requested):

src/pushsource/_impl/backend/koji_source.py
rohanpm: I agree on the "is not None", fixed it.

I don't think it would be right to make it just "if ok_names and ...".  The reason is that it loses the distinction between "caller didn't request any filtering" and "caller requested filtering to an empty list".  Asking to filter to an empty list might seem weird since you're always going to get nothing back, but in the long run I think treating an empty list differently to lists of any other length will end up more confusing.

rbikar (approved):

PR #131, opened 2021-06-15
errata: support filtering RPMs by arch

rohanpm:
Add a parameter which allows adjusting the set of arches included in
returned RPMs.

The primary use-case for this at the moment is during alt-src pushes.
In that case, we know that only source RPMs are of interest, and
querying for other RPM arches only costs extra time.


rohanpm:
It can be seen from the build logs that the "travis-ci/push" failure was bogus, it can be safely ignored.

chandwanitulsi (approved):

PR #136, opened 2021-06-17
docs: add an attribute index onto push items

rohanpm:
Previously, in the docs for push item classes, inherited attributes
were not displayed. This is annoying since it means one needs to look at
the base class and subclass docs together to understand what's
available. In extreme cases such as ProductIdPushItem which don't have
any of their own attributes, it gives the impression of a completely
empty class.

Unfortunately none of the built-in sphinx autodoc features seem able to
solve this.  There is an :inherited-members: option, but if enabled it
copies the entire doc strings from the base classes without any
indication of what's inherited and what's not, which is far too
verbose.

So let's add a little extension ourselves which generates a list under
the class description.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -50,6 +50,7 @@ extensions = [
     "sphinx.ext.viewcode",
     # own extensions in 'ext' dir
     "attr_types",
+    "attr_index",
 ]
 
 # Add any paths that contain templates here, relative to this directory.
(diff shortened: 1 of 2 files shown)


chandwanitulsi (approved):

rohanpm:
Holding off on merge of this because I found sphinx.ext.autosummary extension after I'd written this, which seems to be similar to this.  I should check if that's already sufficient first.

rohanpm:
I gave autosummary a try but it doesn't seem a good fit.  It seems to require explicitly listing everything you want documented, and I can't get it to generate docs for attrs-based attributes at all after trying for a bit.

PR #139, opened 2021-06-22
Introduce PushItem.build_info attribute

rohanpm:
We've previously had koji build NVRs in 'build' attribute.

It seems there is a use-case for handling (name, version, release)
components separately. For example, in the container world it's
common to reuse the 'version' from a build as a tag in a container
image repo.

To avoid callers having to reinvent NVR parsing everywhere, let's
provide that info directly on the push items. We put it into a
new class, which we may add other build attributes onto in the
future.


rbikar (approved):

rohanpm:
I've screwed up the merge of these stacked PRs a bit.

PR #153, opened 2021-06-30
chore: update pip-compile workflow


rohanpm:
## Automated update

This is an automated update generated by ansible playbooks.
See [the playbook repo](https://url.corp.redhat.com/300a997) for more information.

### update pip-compile workflow

This GitHub Actions workflow will regularly invoke pip-compile over relevant files in this repo and submit pull requests when python dependencies need updating.

### DIFF
--- a/.github/workflows/pip-compile.yml
+++ b/.github/workflows/pip-compile.yml
@@ -1,0 +1,55 @@
+on:
+  workflow_dispatch: {}
+  schedule:
+  - cron: "20 20 * * 0"
+
+name: "pip-compile: create PR"
+jobs:
+ pip_compile:
+   name: pip-compile
+   runs-on: ubuntu-latest
+   steps:
+     - name: Setup Python
+       uses: actions/setup-python@v2
+       with:
+         # TODO: multi-version support
+         python-version: 3.8
+
+     - name: Install system dependencies
+       run: sudo apt-get install -y libkrb5-dev
+     - name: pip-compile
+       uses: technote-space/create-pr-action@v2
+       with:
+         EXECUTE_COMMANDS: |
+           pip install pip-tools
+           pip-compile -U --generate-hashes test-requirements.in
+           
+         COMMIT_MESSAGE: 'chore: scheduled pip-compile'
+         COMMIT_NAME: 'GitHub Actions'
+         COMMIT_EMAIL: 'noreply@github.com'
+         PR_BRANCH_PREFIX: deps/
+         PR_BRANCH_NAME: 'pip-compile'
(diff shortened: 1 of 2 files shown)


rohanpm:
This is a bit experimental in nature. I'm trying this out in two repos, please see the other repo for a more complete explanation - https://github.com/release-engineering/exodus-gw/pull/345#issuecomment-871136105

I want to try it for this repo in particular because dependabot has already been broken here for months, due to the rpm-py-installer issue.  This change ought to be able to get this repo back on track with dependencies updated regularly.

rbikar (approved):

.github/workflows/pip-compile.yml
```diff
@@ -0,0 +1,55 @@
...
+
+
+     - name: Look up pull request
+       uses: juliangruber/find-pull-request-action@v1
+       id: find-pull-request
+       with:
+         branch: deps/pip-compile
+
+     # Immediately approve it if we can so it can be merged later
+     - name: Approve Pull Request
```
chandwanitulsi, line 48: Would it be okay if the PR is not checked by any of the members and it gets merged automatically?

rohanpm: I think it should be OK. Here's my thoughts on it:

- We're moving towards every repo having all the dependencies pinned and then updated in some way, which generates some work.  Even if we cut it down to (one PR per repo per week) I don't think it's realistic to expect people to do manual review on all of them.
- There are definitely some risks associated with automerge, namely that we can end up pulling in a bad dependency (something malicious, or just a dependency with a showstopper bug which managed to get past testing).
- The delays built-in to the workflow are meant to mitigate the risks.  First, we only look for new updates once per week, so we're not updating as soon as new releases come out.  Second, the automerge happens a couple of days after the change was created rather than instantly.  I can't point to any data, but I expect "package must have remained on pypi for a few days" to help cut down on the risk of something bad getting into the dependency tree of any popular packages; for example if a bad guy manages to hijack some dependency of requests such that "pip install requests" gives you something malicious, I'd be surprised if that goes undetected for more than a day or two.
- And finally, in the case of services I expect an additional (internal) code scanning step to take place after this anyway before we can put container images into production.  (This is a moot point for this particular repo and other libraries because these requirements are only for running the test suite... any kind of hijacking only affects the CI and not the production services where the library is used).

The workflow actually supports both modes (it relies on APPROVAL_TOKEN secret here, if someone doesn't go and set it in the project settings then there's no auto-approval and a human still has to take action.)

chandwanitulsi: These make sense, thanks for the explanation

chandwanitulsi (approved):

PR #156, opened 2021-07-02
pylint: disable unused-private-member

rohanpm:
pylint 2.9.0 (2021-06-29) introduced unused-private-member checker,
but it wrongly flags some of our code as unused (code using
@handles_type decorator in staged backend).


rbikar:
It looks I can't "Add review" to the PR, the page always breaks :( so giving +1 here.

PR #165, opened 2021-07-29
GitHub Actions Workflow

daminijain23:
Added - 
tox-test.yml
release.yml

Removed-
.travis.yml

Updated - tox.ini

https://projects.engineering.redhat.com/browse/RHELDST-7272


.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+      with:
+        python-version: '3.8'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: __token__
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
```
rohanpm, line 24: Could we please use the name PYPI_TOKEN for this instead?  This is more consistent with various other projects already using gh-actions.

.github/workflows/tox-test.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Install Tox
+        run: pip install tox
+      - name: Install pytest cov
+        run: pip install pytest-cov
+      - name: Run Tox
+        run: tox -e cov
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1.0.15
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
rohanpm, line 59: My impression is that a token is not required.

When I submitted the workflow you created into pubtools-pulplib, as I suspected that this token wasn't needed, I deliberately didn't fill in any value in the repo settings.  Codecov does seem to work fine without it, e.g. https://github.com/release-engineering/pubtools-pulplib/pull/119 .  Could we drop this reference to a token?

.github/workflows/tox-test.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install Tox
+        run: pip install tox
+      - name: Run Tox
+        run: |
+          tox -e docs
+          scripts/push-docs
```
rohanpm, line 79: I don't think scripts/push-docs will work here because it is written to look at some Travis environment variables.

Could you please check what I did on https://github.com/release-engineering/pubtools-pulplib/pull/117 and do similar here?

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,py36,static,docs
+envlist = py27,py36,static,docs,py38
```
rohanpm, line 2: I suggest to change py36 into py38 instead (to make it clear we don't want to support py36).

```suggestion
envlist = py27,py38,static,docs
```

rohanpm (changes requested):
@daminijain23 as well as the inline comments, could you please squash some or all of these commits to make this mergeable with a cleaner history?  I did that myself when merging the pubtools-pulplib change, but it would be good if you could take care of it.

If you're not familiar with the concept - my suggestion is learning `git rebase -i`, using it on top of the latest master, then "squash" all the commits together into one with a commit message like "migrate from travis to github actions".

rohanpm:
@daminijain23, the actual content of this looks good to me now, but are you able & willing to squash it into fewer commits?

daminijain23:
@rohanpm squashed into fewer commits.

rohanpm:
@daminijain23 thanks,

Just one other thing - you dropped scripts/push-docs but didn't add the replacement (docs.yml), as in example https://github.com/release-engineering/pubtools-pulplib/pull/117 .  Can it please be added?  If it's not in this PR then someone will have to come and add it afterward.

daminijain23:
@rohanpm Added docs.yml

PR #176, opened 2021-09-10
Use https for errata

AvlWx2014:
Update `ErrataSource` to use https rather than http for Errata Tool XML-RPC endpoint requests. 

Historically, the `ErrataSource` constructor allowed calling code to pass URLs with either http or https schemes, and would force http behind the scenes. For compatibility, the `ErrataSource` constructor still allows calling code to use http or https, but now forces https behind the scenes.

### DIFF
--- a/src/pushsource/_impl/helpers.py
+++ b/src/pushsource/_impl/helpers.py
@@ -1,6 +1,9 @@
 import six
 
 
+from six.moves.urllib.parse import urlparse, ParseResult
+
+
 def list_argument(value, retain_none=False):
     """Convert an argument into a list:
 
@@ -85,3 +88,22 @@ def try_bool(value):
     elif value in ["", "0", "false", "no"]:
         return False
     raise ValueError("Expected a boolean, got %s" % repr(value))
+
+
+def force_https(url):
+    """Force ``url`` to use https as its scheme.
+
+    Parameters:
+        url: str
+            A string representation of a URL.
+
+    Returns:
+        str
+            ``url`` with its scheme replaced with https.
+
+    Throws:
+        Any potential error thrown by the urlparse function during parsing.
+    """
+    parsed = urlparse(url)
+    with_https = ParseResult("https", *parsed[1:])
+    return with_https.geturl()
--- a/tests/errata/test_errata_metadata.py
+++ b/tests/errata/test_errata_metadata.py
@@ -17,9 +17,8 @@ def test_errata_typical_metadata(fake_errata_tool):
     items = list(source)
 
     # It should have queried the expected XML-RPC endpoint.
-    # Note that our https was replaced with http, this is expected!
     assert (
-        fake_errata_tool.last_url == "http://errata.example.com/errata/errata_service"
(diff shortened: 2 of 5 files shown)


src/pushsource/_impl/helpers.py
```diff
@@ -1,6 +1,14 @@
+import sys
+
 import six
 
 
+if sys.version_info.major < 3:
```
rohanpm, line 6: In the other file where urlparse is used, the python2 vs python3 topic is handled via importing from six.moves.  Can you please do that here as well (if it works)?

src/pushsource/_impl/helpers.py
```diff
@@ -85,3 +93,22 @@ def try_bool(value):
     elif value in ["", "0", "false", "no"]:
         return False
     raise ValueError("Expected a boolean, got %s" % repr(value))
+
+
+def force_https(url: str) -> str:
```
rohanpm, line 98: You can't use inline type hints because the code has to be compatible with python2.  Currently, this is the cause of at least some of the CI failures.

src/pushsource/_impl/helpers.py
```diff
@@ -85,3 +93,22 @@ def try_bool(value):
...
+    """Force ``url`` to use https as it's scheme.
```
rohanpm, line 99:
```suggestion
    """Force ``url`` to use https as its scheme.
```

("it's" always means "it is" or "it has")

src/pushsource/_impl/helpers.py
```diff
@@ -85,3 +93,22 @@ def try_bool(value):
...
+
+    Parameters:
+        url: str
+            A string representation of a URL.
+
+    Returns:
+        str
+            ``url`` with it's scheme replaced with https.
```
rohanpm, line 107:
```suggestion
            ``url`` with its scheme replaced with https.
```

rohanpm (changes requested):

src/pushsource/_impl/helpers.py
AvlWx2014: Oh wow that was careless of me to leave in there, good catch. I had hope that PyCharm's version compatibility checks would have warned me of this, but I'm becoming increasingly convinced they aren't very robust.

src/pushsource/_impl/helpers.py
AvlWx2014: Replaced with `from six.moves.urllib.parse import urlparse, ParseResult`.

tests/errata/test_errata_metadata.py
```diff
@@ -19,7 +19,7 @@ def test_errata_typical_metadata(fake_errata_tool):
     # It should have queried the expected XML-RPC endpoint.
     # Note that our https was replaced with http, this is expected!
```
rohanpm, line 20: This comment is now outdated, this will probably confuse a casual reader.  Could you please adjust or remove it?  Note: the same comment appears in a few places.

rohanpm (changes requested):

rohanpm (approved):

PR #179, opened 2021-09-27
Added registry_source

midnightercz:
New source type RegistrySource is added to support pushing
containers from external registries


rohanpm (changes requested):
I'll have more detailed comments later, but the biggest issue first: why use skopeo?

I think a dependency between this library and skopeo, and also the complexity of running it in a container, should not be introduced lightly.  As far as I can tell there's no strong need for it.  The only data which is used from the output of 'skopeo inspect' seems to be these:

### Labels

If you already have code for interacting with a registry (which you do) then it can be near-trivially obtained by just one more request after getting the v2 manifest - as in example:

```
$ curl -H 'Accept: application/vnd.docker.distribution.manifest.v2+json' -L https://registry.access.redhat.com/v2/ubi8/ubi/manifests/8.4-211
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
   "config": {
      "mediaType": "application/vnd.docker.container.image.v1+json",
      "size": 4366,
      "digest": "sha256:53ce4390f2adb1681eb1a90ec8b48c49c015e0a8d336c197637e7f65e365fa9e"
   },
   "layers": [
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 83368932,
         "digest": "sha256:262268b65bd5f33784d6a61514964887bc18bc00c60c588bc62bfae7edca46f1"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 1795,
         "digest": "sha256:06038631a24a25348b51d1bfc7d0a0ee555552a8998f8328f9b657d02dd4c64c"
      }
   ]
}

# Get the 'config' blob referenced above
$ curl -Ls https://registry.access.redhat.com/v2/ubi8/ubi/blobs/sha256:53ce4390f2adb1681eb1a90ec8b48c49c015e0a8d336c197637e7f65e365fa9e | jq .config.Labels
{
  "architecture": "x86_64",
  "build-date": "2021-09-14T16:37:54.438551",
  "com.redhat.build-host": "cpt-1002.osbs.prod.upshift.rdu2.redhat.com",
...
```

### RepoTags

What is this actually?  All tags in the repo pointing at the manifest?

Two points about that:

- I'm not sure it's a good idea to use this field from the skopeo output.  In https://github.com/containers/skopeo/issues/785 we can see several people are complaining about the time the command takes to scan all those tags and asking for a way to disable it, others are suggesting it doesn't make sense for this info to be included in the output at all since it can't be done efficiently using the registry API... I agree with them and wouldn't be surprised if this disappears from later versions of skopeo.  (Is the output even documented and can it be trusted to maintain compatibility?)
- Why should the set of tags pointing at the image affect the result anyway?  For instance using the example of `registry.access.redhat.com/ubi8/ubi:8.4-211`, if someone provides that as a source of content, why do they need to know the set of all tags in that repo pointing at the same image.  We don't do anything similar for other bits of content.  Can we just provide the tag we were asked to operate on?

docs/sources/registry.rst
```diff
@@ -0,0 +1,39 @@
...
+
+* Container images
+
+
+registry URLs
+------------------
+
+Registry source can be obtained by specifying URL:
+
+``registry:registry-hostname/namespace/repository:tag``
```
rohanpm, line 17: I don't think this is accurate for the current implementation.  I also can't find any test which covers this (`Source.get` with a `registry:` URL).

If you want to make it work as documented here, you probably have to rename `registry_uris` to just `url` and/or do some other tricks.

How do you want it to work in the case of more than one image?  If you want e.g. this to work:

`registry:registry.access.redhat.com/ubi7/ubi:latest,registry.access.redhat.com/ubi8/ubi:latest`

... and to produce push items for both the ubi7 and ubi8 images, which seems reasonable to me, I would strongly suggest writing a test for it because I don't think it works right now.

midnightercz:
> I'll have more detailed comments later, but the biggest issue first: why use skopeo?

Because we already use it in pubtools-quay and in the future code will be shared. 

Thanks for pointing out issue with RepoTags, I'll change the code to not use it

rohanpm:
> > I'll have more detailed comments later, but the biggest issue first: why use skopeo?
> 
> Because we already use it in pubtools-quay and in the future code will be shared.
> 

It's not a good enough reason.

Right now the dependencies of this library are relatively lean.  If you can install this library into a virtualenv then you're probably set up to use it already (playing around with `examples/list-push-items` can demonstrate that).  This makes it easily usable from many environments and I want it to stay that way.

Having a backend which depends on a running docker daemon and/or installed skopeo command is, in my view, a big deal which makes it much less convenient to use.  I think it should only be done if other ways are impractical.  Also, isn't much of the code associated with that already known to be a dead end?  (i.e. won't you want to run skopeo locally once there are no legacy environments in scope?)

I don't think this should be a big deal - right now it just runs "skopeo inspect" in one place, and by my reading, all you have to do to replace that one call to skopeo is to do two HTTP requests with some basic JSON parsing.  That will allow a lot of code to be removed from this PR, as well as simplifying the API and avoiding this library having some new dependencies.

midnightercz:
> It's not a good enough reason.
> 
> Right now the dependencies of this library are relatively lean. If you can install this library into a virtualenv then you're probably set up to use it already (playing around with examples/list-push-items can demonstrate that). This makes it easily usable from many environments and I want it to stay that way.
> 
My Idea was to have single place which holds all the code needed for running skopeo commands for pubtools-quay. I didn't want to have it in pubtools quay as now it shows there are use cases in other projects for the such functionality - like this new backend. So yes, we could use two http requests to achieve same functionality, but that means there will be different code for this project and different code for pubtools-quay. I don't see reason why we should maintain two code paths if we can share it. I meant to add this functionality to pubtools project to make it accessible for all other pubtools-* projects, but since you were against it, I guess ideally in the future we'll create pubtools-containers or something like that to gather the functionality.
For the mentioned soon-to-be-deprecated functionality for containers, SkopeoContainerExecutor can be easily replaced with new LocalContainerExecutor which will be just new aggregation of skopeo commands and local executor.

rohanpm:
> we could use two http requests to achieve same functionality, but that means there will be different code for this project and different code for pubtools-quay

I'm not sure that's correct.  You're now implementing a way for pushsource library to get container push item metadata and I assume the intent is to start using this in pubtools-quay, which is good.  But if pubtools-quay uses pushsource to get info on container images, then why does it *also* need to use skopeo inspect separately?  This seems like a design bug as there should ideally be just one source of the metadata.

In concrete terms, I'm looking at latest pubtools-quay code now and skopeo inspect is run in just one place:

```python
        # Check arch if the image is V2S2 manifest
        if manifest["mediaType"] == TagDocker.MANIFEST_V2S2_TYPE:
            arch = executor.skopeo_inspect(reference)["Architecture"]
            # Arch check is not a great way to verify that this is a source image, but there are
            # no better options without having build details
            if arch != "amd64":
                raise BadPushItem(
                    "Image {0} has V2S2 manifest and contains an architecture {1}. Only source "
                    "images are supported, which have arch 'amd64'.".format(reference, arch)
                )
```

But isn't this use-case (figure out whether it's a source image) already part of what you're implementing here in pushsource?  So once pushsource supports it, why would you *also* want to keep this code in pubtools-quay and have two ways of getting container image metadata.

> I don't see reason why we should maintain two code paths if we can share it

Sharing code is good, but it doesn't override all other considerations.  Compared to doing a couple of HTTP requests, running skopeo inspect seems worse in many ways: it adds new dependencies, it'll be slower since it's doing more work, it'll be less reliable as long as it works via containers due to more points of failure, it's harder for us to debug, it complicates the code, it complicates the API as you need to make all the details around running skopeo configurable (which is something the users of this library will also have to deal with).  That's a lot of disadvantages.  If you decided that these are all acceptable for pubtools-quay, OK, but I don't agree for pushsource.

src/pushsource/_impl/backend/registry_source.py
```diff
@@ -0,0 +1,171 @@
...
+        docker_url="unix://var/run/docker.sock",
+        executor_container_image=None,
+        docker_timeout=None,
+        docker_verify_tls=None,
+        docker_cert_path=None,
+    ):
+        """Create a new source.
+
+        Parameters:
+            dest (str, list[str])
```
rohanpm, line 48: The docs don't spell out how "dest" and "destination_tag" on the URIs are supposed to work together, but I'll assume the code as written now is as intended.  What's implemented seems to be that "dest" should contain repos, and "destination_tag" should contain tags, and then the final value of "dest" on each generated item is equal to the repos and tags combined.

I think there are some issues with this:

- It's inconsistent with the handling of 'dest' argument for other sources when dealing with container images.  Here's an example taken from [internal docs](https://docs.engineering.redhat.com/x/60QiDQ): `brew:container_build=ubi8-container-8.5-145&dest=ubi/ubi8:latest,ubi/ubi8:8.5` .  If this backend has an argument named "dest", I think it should use the repo:tag in a consistent manner as the existing container-supporting backend.
- Due to the way that the repos in 'dest' and the tags in the URIs are combined together, we end up with the behavior that each individual image referenced can be sent to a different *tag*, but all of them have to be delivered to the same set of *repos*.  Does this meet the use-case actually?  From the point of view of the caller, this seems arbitrary (why can I set tag on a per-image basis, but can only set repos for all images at once).

The cleanest design is not obvious.  How do you expect to calculate the dest in practice, when this is called?

I suggest initially being consistent with koji source, i.e. let it accept a "dest" and just blindly apply this onto all push items, but in practice the non-trivial callers would be expected to set the dest themselves on the generated items.

src/pushsource/_impl/backend/registry_source.py
```diff
@@ -0,0 +1,171 @@
...
+
+            docker_verify_tls (bool):
+                Whether to perform TLS verification with the Docker client. Disabled by default.
+
+            docker_cert_path (str):
+                Path to Docker certificates for TLS authentication. '~/.docker' by default.
+        """
+        self._registry_uris = registry_uris
+        self._dest = list_argument(dest)
+        self.signing_keys = list_argument(signing_key)
```
rohanpm, line 81: Could you please name all these attributes with a leading `_` to signal that they're private to this class?

tests/docs/test_autodoc_coverage.py
```diff
@@ -27,7 +28,11 @@ def all_rst_content():
 
 def public_class_names():
     all_attrs = dir(pushsource)
-    return [elem for elem in all_attrs if not elem.startswith("_")]
```
rohanpm, line 30: You probably won't need this change after the import of 'util' module is removed.

src/pushsource/__init__.py
```diff
@@ -1,3 +1,5 @@
+from pushsource._impl import utils
```
rohanpm, line 1: Please don't import this here as it doesn't contain any public API.

src/pushsource/_impl/backend/registry_source.py
```diff
@@ -0,0 +1,171 @@
...
+        pass
+
+    def _push_item_from_registry_uri(self, uri, signing_key):
+        schema, rest = uri.split("://")
+        host, rest = rest.split("/", 1)
+        repo, tags_part = rest.split(":", 1)
+        tags = tags_part.split(":")
+        source_tag = tags[0]
+        dest_tags = tags[1:]
+        if not dest_tags:
```
rohanpm, line 103: There is another comment above about "dest", but putting that aside for the moment and just focusing on this one...

I don't think this should be mandatory.  That prevents the caller from making decisions about the destination based on the metadata returned.  If it's mandatory, the caller has to decide up front which destination tag they want on the image when they don't yet know anything about the image except for the URL.  It means the caller of pushsource can't implement some basic things like "use a different tag for SourceContainerImagePushItem vs ContainerImagePushItem".

More generally, it's useful to be able to use these backends without *any* destination involved, e.g. koji source allows this.

src/pushsource/_impl/backend/registry_source.py
```diff
@@ -0,0 +1,171 @@
...
+                The destination(s) to fill in for push items created by this
+                source. If omitted, all push items have empty destinations.
+
+            registry_uris (list[str])
+                List of references to container images with tags+dest tags
+                Format <host>/<namespace>/<repo>:<tag>:<destination_tag>:<destination_tag>
+                Example: registry.redhat.io/ubi:8:latest:8:8.1
+
+            signing_key (list[str])
+                GPG signing key ID(s). If provided, content must be signed
```
rohanpm, line 58: This is not accurate, is it?  It looks like it was copy-pasted from koji backend, but in that backend it means "if we find the content and it's not signed with this key, it's a fatal error".  But here it instead just means "The value you pass will be copied onto the output push items".

Could it please be reworded for accuracy?

Also, I think it should be named dest_signing_key (matching exactly the related attr on the push items).  This makes the difference with KojiSource signing_key clearer, and also avoids a clash if we ever want to actually verify signing keys here similarly as the koji case.

docs/sources/registry.rst
```diff
@@ -0,0 +1,39 @@
...
+
+* Container images
+
+
+registry URLs
+------------------
+
+Registry source can be obtained by specifying URL + dest tags:
+
+``registry:registry-hostname/namespace/repository:tag:dest-tag1:dest-tag2``
```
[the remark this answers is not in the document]
midnightercz: Right, I forgot to update the docs. Idea is to call it like this:
Source.get("registry:", registry_uris=['host/repo/name:src-tag:dest-tag'])
If there's requirement to be able to specify everything in url, I think I'll choose to accept url like this:
registry:images=host/repo/name:tag:dest-tag,host2/repo2/name:tag:dest-tag
to allow specifying multiple hosts

src/pushsource/_impl/backend/registry_source.py
midnightercz: I don't understand why would user want to have different destination tags for source container vs normal container. User has to know (if that matters to her/him) if image is source container or not because you specify src tags on the input. So one repository can contain both source containers and normal containers but just tagged with different tags.
I don't understand the use case when user wants to decide mapping of src to dst tags after initialization of the source

rohanpm: That was an example meant to illustrate that clients might need to set the destination based on info not available up-front, for future use-cases that you and I currently aren't aware of.  It's unnecessarily restrictive, and inconsistent with other sources, to require the caller to specify a destination before you'll let them get any info about the images.

If you want a real use-case which exists today: As a developer using this library, I want to be able to see realistic examples of what this source will return for an image without actually pushing anything so that I can better understand the data produced by this library, and I don't want to be forced to provide some "dummy" destination in order to do that.  That's what I meant by "it's useful to be able to use them without a destination".

src/pushsource/_impl/backend/registry_source.py
```diff
@@ -0,0 +1,143 @@
...
+    def __init__(
+        self,
+        images_str,
+        dest_repos=None,
+        dest_signing_key=None,
+    ):
+        """Create a new source.
+
+        Parameters:
+            dest_repos str,
```
rohanpm, line 37: I don't think this and the destination_tag stuff should be included, it still seems arbitrary and overly specialized for your current use-case (by the way, is prototype code using this available already? Can you share if so?)

If we compare all the backends now:

- errata: the underlying data source (ET) knows the destination, so it's filled in automatically.
- staged: the underlying data source (filesystem) knows the destination, so it's filled in automatically.
- koji: the underlying data source does not know the destination. There is one parameter, 'dest', which can be used to fill in the destination if all push items are going to the same place. The resulting 'dest' in the output is an arbitrary string, the source does not require it to be any particular format or represent any specific thing.  For any more complex cases, the caller has to set the destination after getting the push items.
- registry: the underlying data source does not know the destination. There are two parameters, 'dest_repos' and 'images_str', influencing the destination; it'll be set on all push items by combining data from these parameters in  such a way that dest_repos is applied to *all* push items and different tags from images_str will be applied to *each* push item.  The resulting 'dest' in the output is specifically a container image repo:tag.  For any more complex cases, the caller has to pass a fake destination, and can then rewrite the destination after getting the push items.

koji and registry shouldn't work so differently, and registry shouldn't assume that we're always gathering push items for push to another registry (which seems implied by the repo:tag output format).

I think "if in doubt, leave it out" applies here.  Can you please either make registry work exactly like koji, or even drop the destination handling entirely and require it to be handled by the caller?

rohanpm (changes requested):

src/pushsource/_impl/backend/registry_source.py
midnightercz: yes, for koji source tags are externally provided. Also that's why it's not possible to map exact source tags to exact destination tags. You can only set destination tags for all push items. I don't find it very useful, that's why I implemented it this way. 
Or do you have another idea how to do the mapping?

rohanpm:
Still crashes, although differently from before:

```python
(python3) [rmcgover@photinia pushsource]$ git rev-parse HEAD
df997d31c54500f2aec70ee8161aed5031483f21
(python3) [rmcgover@photinia pushsource]$ examples/list-push-items 'registry:image=registry.access.redhat.com/ubi8-minimal:8.4-210'
# Loaded source registry:image=registry.access.redhat.com/ubi8-minimal:8.4-210
Traceback (most recent call last):
  File "examples/list-push-items", line 97, in <module>
    main()
  File "examples/list-push-items", line 93, in main
    run(p)
  File "examples/list-push-items", line 61, in run
    for pushitem in source:
  File "/home/rmcgover/src/pushsource/src/pushsource/_impl/backend/registry_source.py", line 127, in __iter__
    yield self._push_item_from_registry_uri(uri, key)
  File "/home/rmcgover/src/pushsource/src/pushsource/_impl/backend/registry_source.py", line 94, in _push_item_from_registry_uri
    ContainerImageDigestPullSpec(
  File "<attrs generated init pushsource._impl.model.container.ContainerImageDigestPullSpec>", line 11, in __init__
  File "/home/rmcgover/dev/python3/lib/python3.8/site-packages/attr/validators.py", line 35, in __call__
    raise TypeError(
TypeError: ("'digest' must be (<class 'str'>,) (got None that is a <class 'NoneType'>).", Attribute(name='digest', default=NOTHING, validator=<instance_of validator for type (<class 'str'>,)>, repr=True, eq=True, eq_key=None, order=True, order_key=None, hash=None, init=True, metadata=mappingproxy({}), type=<class 'str'>, converter=None, kw_only=True, inherited=False, on_setattr=None), (<class 'str'>,), None)
```

It should also be fixed that the tests manage to pass despite the crash.

tests/test_containers.py
```diff
@@ -0,0 +1,710 @@
...
+)
+
+from pushsource._impl.utils.containers.request import (
+    parse_401_response_headers,
+    AuthToken,
+)
+
+
+@pytest.fixture
+def fake_home():
```
rohanpm, line 44: This function can be simplified by:

- using tmpdir fixture rather than using mkdtemp - then pytest will clean up the directory automatically
- using monkeypatch.setenv - then pytest will undo the env changes automatically

It would look something like:

```python
@pytest.fixture(autouse=True)
def fake_home(tmpdir, monkeypatch):
    home = tmpdir.mkdir("home")
    monkeypatch.setenv("HOME", str(home))
    home.mkdir(".docker").join("config.json").write(...<desired content>)
```

Up to you, but a lot of the current code is reinventing what pytest can already do.

tox.ini
```diff
@@ -27,7 +28,7 @@ commands=pidiff pushsource .
 [testenv:cov]
 usedevelop=true
 commands=
-	pytest --cov-report=html --cov=src {posargs}
+	pytest --cov-report=html --cov=src --cov=src/pushsource/_impl/utils ./ src/pushsource/_impl/utils {posargs}
```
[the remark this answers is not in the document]
rohanpm: Can you please undo this change:

- it shouldn't be needed
- if it was needed, it would not be correct to make this change only for "cov".  You aren't only changing coverage-related arguments here, you're also asking it to look for tests under src/pushsource/_impl/utils and it wouldn't make sense to do that only for one testenv.

rohanpm (changes requested):
Thanks, except for the sha255 issue this now seems to work and the behavior looks OK to me:

```
$ examples/list-push-items 'registry:image=registry.access.redhat.com/ubi8-minimal:8.4-210'
# Loaded source registry:image=registry.access.redhat.com/ubi8-minimal:8.4-210
ContainerImagePushItem(
    name="registry.access.redhat.com/ubi8-minimal:8.4-210",
    state="PENDING",
    src="registry.access.redhat.com/ubi8-minimal:8.4-210",
    dest=[],
    md5sum=None,
    sha256sum=None,
    origin=None,
    build=None,
    build_info=None,
    signing_key=None,
    dest_signing_key=None,
    source_tags=["8.4-210"],
    labels=frozendict({}),
    arch=None,
    pull_info=ContainerImagePullInfo(
        tag_specs=[
            ContainerImageTagPullSpec(
                registry="registry.access.redhat.com",
                repository="ubi8-minimal",
                tag="8.4-210",
                media_types=[
                    "application/vnd.docker.distribution.manifest.list.v2+json"
                ],
            )
        ],
        digest_specs=[
            ContainerImageDigestPullSpec(
                registry="registry.access.redhat.com",
                repository="ubi8-minimal",
                digest="sha255:54ef2173bba7384dc7609e8affbae1c36f8a3ec137cacc0866116d65dd4b9afe",
                media_type="application/vnd.docker.distribution.manifest.list.v2+json",
            )
        ],
        media_types=["application/vnd.docker.distribution.manifest.list.v2+json"],
    ),
),
# 1 item(s) found in source
```

Just a few more points:

1. Asking for an image by digest doesn't work (e.g. registry:image=registry.access.redhat.com/ubi8-minimal@sha256:54ef2173bba7384dc7609e8affbae1c36f8a3ec137cacc0866116d65dd4b9afe).  If not currently needed then I don't think this has to work, but would it be worth raising a proper error message here?  For example, raising ValueError if any of "image" contain the "@" character making it immediately obvious this is not supported?  Right now it queries an invalid URL and crashes with a 404 error.

2. The output (at least on a list) ends up containing an empty value for 'labels'.  Could you please confirm if that's OK, as there does seem to be code attempting to deal with labels.

3. After the rest of the issues are addressed could you please rebase the PR and squash to a smaller number of commits, or even just one?

rohanpm:
For the pylint failures like these:

```
pushsource/_impl/utils/containers/request.py:65:8: W0707: Consider explicitly re-raising using the 'from' keyword (raise-missing-from)
```

Although 'raise from' is not available on python2, I looked into this and found that if you use six.raise_from, it fixes the pylint error (as well as improving debuggability on python3).

Could you please fix those places to use six.raise_from?  That should be enough to get CI fully passing.

PR #181, opened 2021-09-29
Refresh dependency management & pip-compile workflow

rohanpm:
This commit updates the pip-compile setup for this repo, bringing it
in line with latest conventions across our projects. Changes are mostly
deployed from templates in ansible.

Changes in detail include:

1. make pip-compile workflow use a configurable token
2. use pip-compile for both requirements and test-requirements
3. share test-requirements.txt with more testenvs, rather than having
   dependencies listed in tox.ini

Point (1) is the primary motivation for doing this now. CI in this repo
was recently migrated from Travis to Github Actions, and this doesn't
work properly if the pip-compile workflow uses the default token
provided by actions, because that token isn't allowed to trigger
additional workflows.


rohanpm:
See https://github.com/release-engineering/pushsource/pull/180 as an example of a pip-compile PR which is not getting tested properly at the moment.

nathanegillett (approved):

rajulkumar (approved):

PR #183, opened 2021-10-07
Mitigate spurious TimeoutError for slow caller [RHELDST-6064]

rbikar:
Mitigate spurious TimeoutError for slow caller [RHELDST-6064]

Previously iterating over any *Source object suffered from
spurious TimeoutError when caller was slow. The generator 'as_completed'
is unable to reset or stop timer for timeout, the timer for timeout
starts in original call to this generator.

Instead an alternative generator called 'as_completed_with_timeout_reset'
was implemented in helpers.py. It provides following functionality:
- yields finished futures as soon as any of them is finished
- after a finished future is yielded, there is no timeout check in
  place after yield
- timeout is only applied when next item from generator is requested

This approach allows caller to work as long as it needs, TimeoutError
is raised only in case when next requested item from pushsource
takes too long to yield to caller.

Fixed a couple of imports in order to pass pylint checks.

### DIFF
--- a/src/pushsource/_impl/helpers.py
+++ b/src/pushsource/_impl/helpers.py
@@ -142,9 +142,6 @@ def as_completed_with_timeout_reset(futures, timeout):
         finished, pending = wait(pending, timeout=timeout, return_when=FIRST_COMPLETED)
         # no future finished in timeout given, raise TimeoutError
         if not finished:
-            # try to cancel futures, we can't cancel those already running or finished
-            for fs in pending:
-                fs.cancel()
             raise TimeoutError(
                 "%d (of %d) futures unfinished" % (len(pending), total_futures)
             )
--- a/src/pushsource/_impl/helpers.py
+++ b/src/pushsource/_impl/helpers.py
@@ -1,5 +1,10 @@
+from concurrent.futures import wait, FIRST_COMPLETED
+
 import six
 
+# in py2 TimeoutError is not built-in exception
+if six.PY2:  # pragma: no cover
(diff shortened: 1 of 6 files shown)


src/pushsource/_impl/helpers.py
```diff
@@ -107,3 +112,42 @@ def force_https(url):
...
+        TimeoutError, if waiting for some Future to finish in wait() call takes more
+        time than given timeout.
+    """
+    total_futures = len(futures)
+    pending = futures
+    while pending:
+        finished, pending = wait(pending, timeout=timeout, return_when=FIRST_COMPLETED)
+        # no future finished in timeout given, raise TimeoutError
+        if not finished:
+            # try to cancel futures, we can't cancel those already running or finished
```
rohanpm, line 145: I think this cancel part doesn't fit, could we please drop it?  The rest looks good.

The reasons cancel doesn't seem right here are:
- the standard as_completed doesn't do this (and naming of this customized version doesn't give any hint there's a difference)
- it means the caller can't possibly decide to retry and wait for a bit longer on the current futures, because we're effectively "breaking" those futures before we return.
- it shouldn't be needed, because we should already be setting up our executors to cancel on shutdown.  So, if this raised TimeoutError here ends up being fatal, then we're already going to cancel all pending futures before allowing the exception to propagate to the caller of pushsource.
- unless I missed something, it wasn't covered by the tests.

rbikar: OK, I'll drop it.

rohanpm (approved):

PR #208, opened 2022-01-18
Fixed inspecting v1 images

midnightercz:
Unify inspected output for v1 images. Also tolarted 404 when getting
digest of manifest for containers which have only one type of the
manifests

### DIFF
--- a/src/pushsource/_impl/backend/registry_source.py
+++ b/src/pushsource/_impl/backend/registry_source.py
@@ -111,6 +111,7 @@ class RegistrySource(Source):
                 except requests.exceptions.HTTPError as e:
                     if e.response.status_code == 404:
                         continue
+                    raise e
 
             manifest_details = self._manifests[source_uri][mtype]
             content_type, _, _ = manifest_details
--- a/src/pushsource/_impl/backend/registry_source.py
+++ b/src/pushsource/_impl/backend/registry_source.py
@@ -1,6 +1,8 @@
 import re
 import logging
 
+import requests
+
 from ..source import Source
 from ..model import (
     ContainerImagePushItem,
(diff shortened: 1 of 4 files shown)


src/pushsource/_impl/backend/registry_source.py
```diff
@@ -98,13 +100,17 @@ def _push_item_from_registry_uri(self, uri, signing_key):
...
+                    manifest_details = get_manifest(
+                        "%s://%s" % (schema, host),
+                        repo,
+                        src_tag,
+                        manifest_types=[mtype],
+                    )
+                    self._manifests[source_uri][mtype] = manifest_details
+                except requests.exceptions.HTTPError as e:
+                    if e.response.status_code == 404:
+                        continue
```
rohanpm, line 113: Did you want to re-raise if it was something other than a 404?

Otherwise it seems like we can just crash at the next line trying to access `self._manifests[source_uri][mtype]`.

src/pushsource/_impl/utils/containers/request.py
```diff
@@ -302,11 +302,17 @@ def inspect(registry, repo, digest, token=None):
     token = token or AuthToken()
 
     manifest_type, digest, manifest = get_manifest(
-        registry, repo, digest, manifest_types=[MEDIATYPE_SCHEMA2_V2_LIST], token=token
+        registry, repo, digest, manifest_types=[MEDIATYPE_SCHEMA2_V2], token=token
     )
     if manifest_type == MEDIATYPE_SCHEMA2_V2:
+        print("v2 sch2")
```
rohanpm, line 308: Please remove the prints.  You could replace them with some debug logs if the info might be useful.

rohanpm (changes requested):

PR #209, opened 2022-01-24
tests: fix crash on frozendict>=2.2

rohanpm:
As of frozendict 2.2.0, isinstance(frozendict(), dict) is no longer
true. This breaks the YAML serialization used in this test. Although
frozendict advertises itself as a collections.abc.Mapping, that is not
enough for PyYAML to work.

Fix it by converting all mappings to plain old dicts during
serialization.

### DIFF
--- a/tests/baseline/test_baseline.py
+++ b/tests/baseline/test_baseline.py
@@ -10,6 +10,8 @@ import attr
 import jinja2
 import pytest
 from mock import patch
+from six.moves.collections_abc import Mapping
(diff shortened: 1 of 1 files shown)


rohanpm:
This is supposed to fix the failing CI on https://github.com/release-engineering/pushsource/pull/206 which was triggered by Marco-Sulla/python-frozendict#47 .

rbikar (approved):

PR #213, opened 2022-02-02
Refresh integration with black, drop from 'tox -e static'

rohanpm:
black was previously integrated into both tox and pre-commit.
The version used by tox would be managed by pip-compile.

This causes some issues: if our pip-compile github action updates
to a new version of black which adjusts any formatting, this will cause
CI to fail as it does not automatically run black over all files and
include changes in the commit.

pre-commit.ci on the other hand does support doing exactly that.
Therefore, for black and similar "auto-fixing" tools, it seems better to
leave them solely to pre-commit and not try to include them in tox.

As part of this change, the pre-commit config for black was updated to
ensure pre-commit autoupdates will succeed from this point onwards.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
 -   repo: https://github.com/ambv/black
-    rev: 026c81b83454f176a9f9253cbfb70be2c159d822
+    rev: 22.1.0
     hooks:
     - id: black
-      language_version: python3.6
+      language_version: python3.9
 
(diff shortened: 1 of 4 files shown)


rohanpm:
This is supposed to fix the pip-compile failure in https://github.com/release-engineering/pushsource/pull/21.

negillett (approved):

rajulkumar:
> This is supposed to fix the pip-compile failure in #21.
I think you meant "failure in #211"

rajulkumar (approved):

PR #220, opened 2022-03-07
Add support for "RAW" content via DirectoryPushItem [RHELDST-10120]

crungehottman:
DirectoryPushItem is a PushItem where "src" is the path to a "RAW"
directory, and the "dest" is the directory's name under the staging
area.

For example, provided the following staging directory,

```
/
└── staging
    └── rhel-atomic-host-ostree
        └── RAW
            ├── a-file
            └── a-dir
                └── some-file.txt

```
If pushsource is asked to load "staged:/staging", it will generate
a DirectoryPushItem(src="/staging/rhel-atomic-host-tree/RAW",
dest=["rhel-atomic-host-ostree"]). It will not generate any
FilePushItem objects for the files inside the directory.


src/pushsource/_impl/model/directory.py
```diff
@@ -0,0 +1,7 @@
+from .base import PushItem
+from .. import compat_attr as attr
+
+
+@attr.s()
+class DirectoryPushItem(PushItem):
+    """A :class:`~pushsource.PushItem` representing a directory."""
```
rohanpm, line 7: Minor - this could do with another sentence or two of explanation, e.g. something like:

> On a directory push item, the `src` attribute contains the full path to a directory tree.
> It should generally be interpreted as a request to recursively publish that entire directory
> tree as-is.

docs/model/directory.rst
```diff
@@ -0,0 +1,5 @@
+Push items: directory
+================
```
rohanpm, line 2: I think this will generate a warning as the '=' below the heading doesn't have the right number of characters?

```suggestion
Push items: directory
=====================
```

src/pushsource/_impl/backend/staged/staged_directory.py
```diff
@@ -0,0 +1,18 @@
...
+from ...model import DirectoryPushItem
+from .staged_base import StagedBaseMixin, handles_type
+
+LOG = logging.getLogger("pushsource")
+
+
+class StagedDirectoryMixin(StagedBaseMixin):
+    @handles_type("RAW")
+    # pylint: disable=unused-argument
+    def __directory_push_item(self, leafdir, metadata, entry):
```
rohanpm, line 12: I think this is not right, as the handles_type decorator was designed to be applied onto a method which is called once for each file in the leafdir, not once for the entire leafdir.  That's why my prototype didn't do it this way.

I believe it only seems to work right now in the test because simple_directories/dest1/RAW contains exactly one file, and because `__mixin_push_items` in the base class does not recursively walk the directory.  So it'll see exactly one file and you'll get exactly one DirectoryPushItem back.  If you adjust the testcase so the RAW directory contains more than one file at the top level, I think you'll start to see some duplicate DirectoryPushItem objects yielded.

Also, the case of a completely empty RAW directory should still generate a DirectoryPushItem.

rohanpm (changes requested):
Could you please ensure the docs at https://release-engineering.github.io/pushsource/sources/staged.html are updated too?

docs/sources/staged.rst
```diff
@@ -185,6 +187,17 @@ Files in this directory must have metadata included in ``staged.yaml``.
 
 Will yield instances of :class:`~pushsource.AmiPushItem`.
 
+root/destination/RAW/\*
+.........................
```
negillett, line 191:
```suggestion
root/destination/RAW/\*
.......................
```

negillett (approved):

tests/baseline/cases/staged-simple-directories.yml
```diff
@@ -0,0 +1,46 @@
...
+    build: null
+    build_info: null
+    dest:
+    - dest2
+    md5sum: null
+    name: dest2
+    origin: null
+    sha256sum: null
+    signing_key: null
+    src: {{ src_dir }}/tests/staged/data/simple_directories/dest2/RAW
```
rohanpm, line 33: Since there's no special logic in the code to deal with hidden files or .gitkeep files, this test isn't covering an empty RAW directory as was probably intended.  .gitkeep files aren't special so from our code's point of view, it's just like any other file sitting in that directory.

If you really want to cover the empty directory case, you'll probably have to do something special in tests like creating the directory under tmpdir at the time the test runs.  I don't think it matters so much though, as the way the code works now there's negligible risk that it would break for empty directories.

rohanpm (approved):

dichn (approved):

PR #231, opened 2022-03-31
Improve AMI billing codes handling [RHELDST-10593]

rbikar:
This change makes billing_codes attribute in AMI staged
push item required - but codes inside billing_codes are allowed to
be empty or completely omitted.

This will allow creation of AMI push-item without billing-codes but
we can still use billing_codes.name that comprises AMI image name.

### DIFF
--- a/src/pushsource/_impl/model/ami.py
+++ b/src/pushsource/_impl/model/ami.py
@@ -102,6 +102,6 @@ class AmiPushItem(PushItem):
     billing_codes = attr.ib(
         type=AmiBillingCodes,
         default=None,
-        validator=optional(instance_of(AmiBillingCodes)),
+        validator=instance_of(AmiBillingCodes),
     )
     """Billing codes associated with this image."""
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/psf/black
-    rev: 22.1.0
+    rev: 22.3.0
     hooks:
     - id: black
       language_version: python3.9
--- a/src/pushsource/_impl/schema/staged-schema.yaml
+++ b/src/pushsource/_impl/schema/staged-schema.yaml
@@ -105,6 +105,7 @@ definitions:
         - virtualization
         - volume
         - root_device
+        - billing_codes
 
         additionalProperties: false
 
@@ -226,7 +227,6 @@ definitions:
         # Billing codes for an AMI
         type:
         - object
-        - "null"
         properties:
             name:
                 type: string
@@ -236,7 +236,6 @@ definitions:
                     type: string
         required:
         - name
-        - codes
 
 ###############################################################################
 #  Main schema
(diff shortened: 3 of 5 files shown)


rbikar:
Bumped black version because of:
```
Traceback (most recent call last):
  File "/home/rbikar/.cache/pre-commit/reponvk4t1fu/py_env-python3.9/bin/black", line 8, in <module>
    sys.exit(patched_main())
  File "/home/rbikar/.cache/pre-commit/reponvk4t1fu/py_env-python3.9/lib/python3.9/site-packages/black/__init__.py", line 1423, in patched_main
    patch_click()
  File "/home/rbikar/.cache/pre-commit/reponvk4t1fu/py_env-python3.9/lib/python3.9/site-packages/black/__init__.py", line 1409, in patch_click
    from click import _unicodefun
ImportError: cannot import name '_unicodefun' from 'click' 
```

src/pushsource/_impl/model/ami.py
```diff
@@ -102,6 +102,6 @@ class AmiPushItem(PushItem):
     billing_codes = attr.ib(
         type=AmiBillingCodes,
         default=None,
-        validator=optional(instance_of(AmiBillingCodes)),
+        validator=instance_of(AmiBillingCodes),
```
rohanpm, line 105: I don't think we should do this unless absolutely necessary, since it's not backwards-compatible.  Code and data which worked fine on earlier versions of pushsource will now crash.

Can you achieve the goals without a backwards-incompatible change?  I'm not sure I'm getting the big picture here, but would it help if we leave the validator and schema alone and just change the 'default' to construct an empty AmiBillingCodes?

I've now read the issue and I'm not quite getting why this change is being made at all, as RHELDST-10593 seems to talk about making billing codes *less* required but this commit is making them *more* required.  I certainly agree with what you said there about "imho the billing_code entry seems to be abused a bit" though.  You could consider going as far as deprecating this entire field and introducing better replacements.

rohanpm (changes requested):

src/pushsource/_impl/model/ami.py
rohanpm: From https://github.com/release-engineering/pubtools-ami/pull/34#discussion_r840062375 I think I get what you were going for here.

I don't think any change in pushsource is strictly required here, and if I were working the issue I probably wouldn't bother touching it.  But if you wanted to improve things maybe some new fields really are the better way (just on the attrs class - no changes to schema).  I guess if the list of codes were found in a top-level attribute rather than a sub-object, it would have avoided the problem.  My first stab at new fields would be:

- billing_products: a list, same as billing_codes.codes today, defaults to an empty list. The new name matches exactly the same concept in the AWS APIs.
- billing_name: str, same as billing_codes.name today. Not sure if it should default to None or just one of the commonly used values.  Also maybe there is a better word than "billing" here, but billing doesn't seem all that bad, seeing as how the new proposed value "Marketplace" still somehow relates to the billing (or not) of the image.

rbikar: Also one thing is that billing_codes was always treated as required in pubtools-ami - so from this point of view it's actually a fix I guess :)  ...but anyway I need to get in touch with requester of the change. It seems that currently we probably won't need any change in pubtools-ami or in pushsource. 

Deprecation of this field and adding new fields would be ideal.

rbikar:
After discussion with requester I'm dropping this change as not needed.
I'll file an issue for make billing_codes.[codes,name] as top-level attribute

PR #241, opened 2022-05-10
Adding pushsource support for content gateway

jalam453:
Please review the pushsource support for content gateway and add your review and comments.

### DIFF
--- a/docs/model/cgw.rst
+++ b/docs/model/cgw.rst
@@ -1,0 +1,5 @@
+Push items: cgw
+===============
+
+.. autoclass:: pushsource.CGWPushItem()
+   :members:
--- a/docs/sources/staged.rst
+++ b/docs/sources/staged.rst
@@ -11,6 +11,7 @@ Supported content types:
 * RPMs
 * Advisories
 * Generic files
+* Content Gateway metadata
 * Complete directories
 * Amazon Machine Images (AMIs)
 * comps.xml files (yum repo metadata)
@@ -49,6 +50,7 @@ Here is a brief overview of the structure of a staging directory:
   root/destination/COMPS/*.xml
   root/destination/ERRATA/*
   root/destination/FILES/*
+  root/destination/CGW/*
   root/destination/PRODUCTID/*
   root/destination/MODULEMD/*
   root/destination/RPMS/*.rpm
@@ -149,6 +151,13 @@ Files in this directory must have metadata included in ``staged.yaml``.
 For historical reasons, this directory may also be named ``ISOS``.
 
 Will yield instances of :class:`~pushsource.FilePushItem`.
+
+root/destination/CGW/\*
+.........................
+
+Each file in CGW should be a valid YAML containing a list of definitions for objects in Content Gateway.
+
+Will yield instances of :class:`~pushsource.CGWPushItem`.
 
 root/destination/PRODUCTID/\*
 .............................
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -93,6 +93,8 @@ For detailed information, see the API reference of the associated class.
 +--------------+-----------------------------------------------------------------------------+-------------------------------------+----------------------------------------------------+
 | file         | ``file:/tmp/file-to-push``                                                  | n/a                                 | Obtain a single file-backed item of various types. |
 +--------------+-----------------------------------------------------------------------------+                                     |                                                    |
+| cgw          | ``cgw:/tmp/yaml-file-to-push``                                              |                                     |                                                    |
++--------------+-----------------------------------------------------------------------------+                                     |                                                    |
 | dir          | ``dir:/publish-src?dest=/dest1,/dest2``                                     |                                     |                                                    |
 +--------------+-----------------------------------------------------------------------------+                                     |                                                    |
 | rpm          | ``rpm:Downloads/python3-3.7.5-2.fc31.x86_64.rpm?signing_key=a1b2c3``        |                                     | Each backend of this type accepts any arguments    |
--- a/src/pushsource/_impl/backend/staged/staged_cgw.py
+++ b/src/pushsource/_impl/backend/staged/staged_cgw.py
@@ -1,0 +1,13 @@
+import logging
+from ...model import CGWPushItem
+from .staged_base import StagedBaseMixin, handles_type
+
+LOG = logging.getLogger("pushsource")
+
+
+class StagedCGWMixin(StagedBaseMixin):
+    @handles_type("CGW")
+    def __push_item(self, leafdir, _, entry):
+        return CGWPushItem(
+            name=entry.name, src=entry.path, origin=leafdir.topdir, dest=[leafdir.dest]
+        )
--- a/src/pushsource/_impl/model/cgw.py
+++ b/src/pushsource/_impl/model/cgw.py
@@ -1,0 +1,15 @@
+from .base import PushItem
+from .. import compat_attr as attr
+
+
+@attr.s()
+class CGWPushItem(PushItem):
+    """A :class:`~pushsource.PushItem` representing a Content Gateway (CGW) YAML file.
+
+    For push items of this type, the :meth:`~pushsource.PushItem.src`
+    attribute refers to a YAML file containing list of definitions
+    for objects in Content Gateway.
+
+    This library does not verify that the referenced YAML file is valid.
+
+    """
--- a/tests/baseline/cases/direct-cgw.yml
+++ b/tests/baseline/cases/direct-cgw.yml
@@ -1,0 +1,22 @@
+# A pushsource library testcase.
+#
+# This file was generated from a template.
+# To regenerate, run test_baseline.py with PUSHSOURCE_UPDATE_BASELINE=1.
+
+# URL of Source to test.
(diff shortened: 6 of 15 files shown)


rohanpm (changes requested):
I don't think this really adds anything, because the generated push items are not flagged as being related to CGW in any way.

Could you explain how the user of the library is expected to consume this?  Usually, the pattern is to do something like this:

```python
for item in Source.get(...):
  if isinstance(item, FilePushItem):
    do_something_for_file(item)
  elif isinstance(item, RpmPushItem):
    do_something_for_rpm(item)
  # ... and so on
```

But with this change, you are just generating more FilePushItem with the same fields as the items under FILES directory.  So the caller doesn't seem to get any kind of an indication that a certain item is a CGW definition (if that's what it's supposed to be).  I see no clear reason why the user should put things under CGW directory in the first place rather than simply FILES.

So, two requests:

- can you please update the docs for staged source ( https://release-engineering.github.io/pushsource/sources/staged.html ) to explain what files should be placed under CGW directory?
- can you please show a snippet of code demonstrating how you'd expect these items to be used?

src/pushsource/_impl/model/cgw.py
```diff
@@ -0,0 +1,15 @@
+from .base import PushItem
+from .. import compat_attr as attr
+
+
+@attr.s()
+class CGWPushItem(PushItem):
+    """A :class:`~pushsource.PushItem` representing content gateway(CGW) YAML file.
```
rohanpm, line 7:
```suggestion
    """A :class:`~pushsource.PushItem` representing a Content Gateway (CGW) YAML file.
```

src/pushsource/_impl/model/cgw.py
```diff
@@ -0,0 +1,15 @@
...
+
+    For push items of this type, the :meth:`~pushsource.PushItem.src`
+    attribute refers to a YAML file containing list of definitions
+    for objects in content gateway.
```
rohanpm, line 11:
```suggestion
    for objects in Content Gateway.
```

docs/sources/staged.rst
```diff
@@ -150,6 +152,13 @@ For historical reasons, this directory may also be named ``ISOS``.
 
 Will yield instances of :class:`~pushsource.FilePushItem`.
 
+root/destination/CGW/\*
+.........................
+
+Each file in CGW should be a valid YAML containing list of definitions for objects in content gateway.
```
rohanpm, line 158:
```suggestion
Each file in CGW should be a valid YAML containing a list of definitions for objects in Content Gateway.
```

docs/model/cgw.rst
```diff
@@ -0,0 +1,5 @@
+Push items: cgw
+================
```
rohanpm, line 2: If the underline doesn't match exactly the section title in length, I think it generates a sphinx warning.

```suggestion
===============
```

docs/sources/staged.rst
```diff
@@ -11,6 +11,7 @@ Supported content types:
 * RPMs
 * Advisories
 * Generic files
+* Content Gateway
```
rohanpm, line 14: I think it should mention that we are only dealing with metadata or definitions for CGW, like:

```suggestion
* Content Gateway metadata
```

rohanpm (changes requested):
What's here right now is good and I have only minor suggestions for the docs.

However, can you please make sure that there are tests covering the CGW cases?  At least one test using 'staged' backend with a directory having a CGW subdir, and at least one test using 'cgw' backend.

In both cases you should be able to add a test by dropping a new file into tests/baseline/cases directory, see the existing files there to get an idea how it works.

docs/userguide.rst
```diff
@@ -93,6 +93,8 @@ For detailed information, see the API reference of the associated class.
 +--------------+-----------------------------------------------------------------------------+-------------------------------------+----------------------------------------------------+
 | file         | ``file:/tmp/file-to-push``                                                  | n/a                                 | Obtain a single file-backed item of various types. |
 +--------------+-----------------------------------------------------------------------------+                                     |                                                    |
+| cgw         | ``cgw:/tmp/yaml-file-to-push``                                               |                                     |                                                    |
```
rohanpm, line 96: I think there is a whitespace error here?  Doesn't seem to line up perfectly, although it's not easy to tell in the github UI.

```suggestion
| cgw          | ``cgw:/tmp/yaml-file-to-push``                                               |                                     |                                                    |
```

tests/baseline/cases/direct-cgw.yml
```diff
@@ -0,0 +1,23 @@
+# A pushsource library testcase.
+#
+# This file was generated from a template.
+# To regenerate, run test_baseline.py with PUSHSOURCE_UPDATE_BASELINE=1.
+
+# URL of Source to test.
+url: "CGW:my/great/file.yaml"
+
+# Push items generated from above.
```
rohanpm, line 9: The test fails right now, which makes sense due to some mismatches in the data below.  For example why would the name be custom-filename, or the dest be repo1.

You should be able to reproduce this test failure locally.  Then you should do what the failure message suggests: set environment variable PUSHSOURCE_UPDATE_BASELINE=1 and run the test again.  It should then update this file with the output, i.e. the items you get if you request a source of "CGW:my/great/file.yaml".  If that output looks correct, add the update into your PR.

tests/staged/data/simple_cgw/CGW/staged.yaml
```diff
@@ -0,0 +1,50 @@
+header:
```
rohanpm, line 1: There are a few issues with this:

First, just adding this file doesn't cause it to be used by any test.  You need to adjust or add at least one test which uses it.  The easiest way to do this is probably to copy an existing similar baseline case like tests/baseline/cases/staged-simple-comps.yml and tweak it to point at the simple_cgw directory.

The rest of the complaints I have are about the content of the file.  Now it is true that pushsource library doesn't parse this file in any way, so the content of this file doesn't really have to be correct.  You could put anything at all in here and the tests could still pass.  Right now though the content seems very misleading which makes me think there could be some misunderstandings.  I would rather that the file is either a correct CGW metadata file as will actually be used when doing a push to CGW, or else it's obviously just a placeholder for testing.

The problems are:
- file doesn't have to be named staged.yaml, this is different from the staged.yaml at root of the staging dir (which would be at tests/staged/data/simple_cgw/staged.yaml here - not in CGW directory).  It would be better to name it something else to reduce confusion.
- I don't think the file should have that "header", "payload" structure. It has to be in the top-level staged.yaml file purely for backwards-compatibility reasons.
- The file right now seems to be more like some kind of a schema than a valid CGW metadata file.  For example, the "state|present" string - obviously in reality it has to be one or the other.
- The payload right now has a list with probably more elements than was intended.  For example, right now it has:

```yaml
- type: file
- state: present|absent
- ...
```

So that's a list like: `[{"type": "file"}, {"state": "present|absent"}, ...]`.  But I suspect it was intended to instead write:

```yaml
- type: file
  state: present|absent
  ...
```

Which would give you something quite different: `[{"type": "file", "state": "present|absent", ...}, ...]`  In other words there are probably some '-' in the YAML which should not be there.

Do you have any *real* CGW metadata file which could be used for the test data here?  If not, I'd prefer it if this entire file were just a placeholder with a comment, like:

```yaml
# CGW metadata should go here.
# This library does not parse the file, and this is just a placeholder for testing.
```

rohanpm (changes requested):

tests/staged/data/simple_cgw/CGW/staged.yaml
jalam453: Thanks for the inputs. I have updated the files accordingly.

docs/userguide.rst
```diff
@@ -93,6 +93,8 @@ For detailed information, see the API reference of the associated class.
 +--------------+-----------------------------------------------------------------------------+-------------------------------------+----------------------------------------------------+
 | file         | ``file:/tmp/file-to-push``                                                  | n/a                                 | Obtain a single file-backed item of various types. |
 +--------------+-----------------------------------------------------------------------------+                                     |                                                    |
+| cgw          | ``cgw:/tmp/yaml-file-to-push``                                               |                                     |                                                    |
```
rohanpm, line 96: I think there is still a whitespace error here.  This time in the third and later instances of `|` character.  It comes because my suggestion added one space near the beginning and didn't delete a space from later in the line.

rohanpm (commented):
There was still one tiny error in the docs I think, commented inline.

Apart from that, to finish this off can we please have:

- Rebase & squash the commits in the PR so that there is just one commit with all the changes.
- Get someone from CLOUDDST to Approve this as well.  My approval will consider only the correctness of the code, but I want someone from CLOUDDST to verify whether the changes here actually meet the requirements of the associated epic.

src/pushsource/_impl/backend/staged/staged_cgw.py
```diff
@@ -0,0 +1,15 @@
+import logging
+import os
```
rohanpm, line 2: The os import is not used here.  Seems you need to remove this for CI to pass.

rohanpm (approved):

midnightercz (approved):

PR #250, opened 2022-06-21
Only yield a push item if source file exists [CLOUDDST-13707]

querti:
Sources of push items are generally located in locally mounted NFSs.
Migrating pub to Openshift has introduced a new limitation: it may take
up to 10 minutes to propagate the changes to the mounted volumes. This
means that a source file may not be present during a push, causing
falures. As a workaround, introduce a polling mechanism that checks
whether a push item source file exists before allowing the push item to
be yielded. Item is yielded immediately if source is not defined.
Polling doesn't happen indefinitely, if the item is not present when the
time limit is reached, an error is raised.

Small unrelated changes:
- Some fixtures were moved to the base confest.py to be usable in all
  test modules
- tox.ini was modified to be able to run specified tests instead of the
  whole suite


rajulkumar (approved):

rohanpm (changes requested):
As well as the inline comments, I want to point out this command which works today if you have pip installed this project:

```
$ pushsource-ls koji:https://koji.fedoraproject.org/kojihub?rpm=python3-3.7.5-2.fc31.x86_64.rpm,python3-3.7.5-2.fc31.src.rpm
# Loaded source koji:https://koji.fedoraproject.org/kojihub?rpm=python3-3.7.5-2.fc31.x86_64.rpm,python3-3.7.5-2.fc31.src.rpm
RpmPushItem(
    name="python3-3.7.5-2.fc31.x86_64.rpm",
    state="PENDING",
    src="/mnt/koji/vol/fedora_koji_archive04/packages/python3/3.7.5/2.fc31/x86_64/python3-3.7.5-2.fc31.x86_64.rpm",
    dest=[],
    md5sum=None,
    sha256sum=None,
    origin=None,
    build="python3-3.7.5-2.fc31",
    build_info=KojiBuildInfo(name="python3", version="3.7.5", release="2.fc31"),
    signing_key=None,
    module_build=None,
),

RpmPushItem(
    name="python3-3.7.5-2.fc31.src.rpm",
    state="PENDING",
    src="/mnt/koji/vol/fedora_koji_archive04/packages/python3/3.7.5/2.fc31/src/python3-3.7.5-2.fc31.src.rpm",
    dest=[],
    md5sum=None,
    sha256sum=None,
    origin=None,
    build="python3-3.7.5-2.fc31",
    build_info=KojiBuildInfo(name="python3", version="3.7.5", release="2.fc31"),
    signing_key=None,
    module_build=None,
),
# 2 item(s) found in source
```

Could you please test the change with this or a similar command? At the moment, I think you'll find the change breaks it quite badly (makes it delay for 15 minutes and then crash).  If you accept the inline suggestions, this should work no worse than it did before.

src/pushsource/_impl/source.py
```diff
@@ -29,14 +31,42 @@ class SourceUrlError(ValueError):
...
+        Yields:
+            PushItem:
+                Created push item.
+        """
+        generator = self.__delegate.__iter__()
+        for item in generator:
+            if not hasattr(item, "src") or not item.src:
+                yield item
+            else:
+                for i in range(self.timeout // self.poll_rate):
```
neithere, line 61: maybe call it something like `max_attempts` and reuse below?

src/pushsource/_impl/source.py
```diff
@@ -29,14 +31,42 @@ class SourceUrlError(ValueError):
...
+                    if os.path.exists(item.src):
+                        break
+                    if i == (self.timeout // self.poll_rate) - 1:
+                        raise RuntimeError(
+                            "Push item source {0} is missing".format(item.src)
```
neithere, line 66: may be a good idea to mention the timeout.

src/pushsource/_impl/source.py
```diff
@@ -29,14 +31,42 @@ class SourceUrlError(ValueError):
...
     def __iter__(self):
-        return self.__delegate.__iter__()
+        """
+        Poll until a source file is present before yielding an item.
+
+        This is necessary due to a limitation, where file changes may not immediately propagate
+        to NFS.
+
+        Raises:
+            RuntimeError:
```
neithere, line 50: Is it safe to offer an iterable which may crash mid-iteration and potentially leave the user with a partially completed operation on their side?

querti:
> As well as the inline comments, I want to point out this command which works today if you have pip installed this project:
> 
> ```
> $ pushsource-ls koji:https://koji.fedoraproject.org/kojihub?rpm=python3-3.7.5-2.fc31.x86_64.rpm,python3-3.7.5-2.fc31.src.rpm
> # Loaded source koji:https://koji.fedoraproject.org/kojihub?rpm=python3-3.7.5-2.fc31.x86_64.rpm,python3-3.7.5-2.fc31.src.rpm
> RpmPushItem(
>     name="python3-3.7.5-2.fc31.x86_64.rpm",
>     state="PENDING",
>     src="/mnt/koji/vol/fedora_koji_archive04/packages/python3/3.7.5/2.fc31/x86_64/python3-3.7.5-2.fc31.x86_64.rpm",
>     dest=[],
>     md5sum=None,
>     sha256sum=None,
>     origin=None,
>     build="python3-3.7.5-2.fc31",
>     build_info=KojiBuildInfo(name="python3", version="3.7.5", release="2.fc31"),
>     signing_key=None,
>     module_build=None,
> ),
> 
> RpmPushItem(
>     name="python3-3.7.5-2.fc31.src.rpm",
>     state="PENDING",
>     src="/mnt/koji/vol/fedora_koji_archive04/packages/python3/3.7.5/2.fc31/src/python3-3.7.5-2.fc31.src.rpm",
>     dest=[],
>     md5sum=None,
>     sha256sum=None,
>     origin=None,
>     build="python3-3.7.5-2.fc31",
>     build_info=KojiBuildInfo(name="python3", version="3.7.5", release="2.fc31"),
>     signing_key=None,
>     module_build=None,
> ),
> # 2 item(s) found in source
> ```
> 
> Could you please test the change with this or a similar command? At the moment, I think you'll find the change breaks it quite badly (makes it delay for 15 minutes and then crash). If you accept the inline suggestions, this should work no worse than it did before.

Actually, when I tried the command everything worked fine. I found out that this was due to a bug where the source wrapper only gets applied to sources that don't have __enter__. Should be fixed now.

src/pushsource/_impl/source.py
```diff
@@ -29,14 +31,42 @@ class SourceUrlError(ValueError):
 
 
 class SourceWrapper(object):
+    timeout = 900
```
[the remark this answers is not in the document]
querti: The values are now set by env variables. Though I moved them directly to __iter__ method so that env varialbes set by tests would be read correctly. I also disabled the feature by default.

Anyway, I have a hard time imagining what "In our case it means we set an env var in pub target settings to explicitly enable the timeout we want." will look like. Aren't source and target separate entities? Can target even influence what happens in source code? And do target settings even have the ability to set env variables? Is this some new logic that will have to be implemented somewhere?

src/pushsource/_impl/source.py
```diff
@@ -29,14 +31,42 @@ class SourceUrlError(ValueError):
...
+        generator = self.__delegate.__iter__()
+        for item in generator:
+            if not hasattr(item, "src") or not item.src:
+                yield item
+            else:
+                for i in range(self.timeout // self.poll_rate):
+                    if os.path.exists(item.src):
+                        break
+                    if i == (self.timeout // self.poll_rate) - 1:
+                        raise RuntimeError(
```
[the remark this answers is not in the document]
querti: Changed

src/pushsource/_impl/source.py
querti: Added

src/pushsource/_impl/source.py
querti: Added

src/pushsource/_impl/source.py
querti: Behavior changed

src/pushsource/_impl/source.py
```diff
@@ -32,25 +37,66 @@ class SourceWrapper(object):
...
+                        "Waiting for %s seconds for source file %s to appear",
+                        poll_rate,
+                        item.src,
+                    )
+                    time.sleep(poll_rate)
+                yield item
 
     def __enter__(self):
+        if hasattr(self.__delegate, "__enter__"):
+            return self.__delegate.__enter__()
```
rohanpm, line 94: I don't think this is right because this causes the caller to escape the SourceWrapper.  Doesn't it?

e.g. if caller does:

```
with Source.get(...) as x:
  for item in x:
    ...
```

If the underlying source implements \_\_enter\_\_ then x is not going to be a SourceWrapper any more, it's going to be whatever the delegate's \_\_enter\_\_ returned.  So the new logic you added in iter won't work.

I think calling enter on the delegate is OK but you still need this to return self.

rohanpm (changes requested):

src/pushsource/_impl/source.py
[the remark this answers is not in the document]
rohanpm: A target is just a bunch of metadata.  Many Pub tasks use some of that metadata to adjust arguments or environment variables at the beginning of the task.

Target settings have the ability to set env vars, it was implemented a while ago in RHELDST-7231.  You were reviewer on that, so if you want to improve awareness of what's happening in the code you could consider putting more effort on keeping up with review invites.

src/pushsource/_impl/source.py
querti: Changed

rohanpm (approved):

PR #253, opened 2022-06-28
Also poll for koji paths

querti:
Koji source checks if source paths exist and raises and exception if
they don't. Add a polling mechanism for this part of the code,
because otherwise the push items will fail to be created before the main
polling mechanism is even attempted.

Also fix a bug where polling is performed 1 less time then expected.

Refers to CLOUDDST-13707


midnightercz (changes requested):

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -283,18 +284,38 @@ def _push_items_from_rpm_meta(self, rpm, meta):
...
+            for key in self._signing_key:
+                if key:
+                    key = key.lower()
+                    candidate = os.path.join(
+                        build_path, self._pathinfo.signed(meta, key)
+                    )
+                else:
+                    candidate = unsigned_path
+                if candidate not in candidate_paths:
+                    candidate_paths.append(candidate)
```
rohanpm, line 305: Since this is in a loop and candidate_paths is a list, aren't we putting duplicates in the list now?  It'll at least mess up the LOG.info below.

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -283,18 +284,38 @@ def _push_items_from_rpm_meta(self, rpm, meta):
...
+                    break
+
+            if rpm_path or i == max_attempts:
+                break
+            LOG.info(
+                "Waiting for %s seconds for any of the following paths to appear: %s",
+                poll_rate,
+                ", ".join(candidate_paths),
+            )
+            time.sleep(poll_rate)
```
rohanpm, line 318: I'm not sure the implemented logic is safe unfortunately.  Depending on exactly how the delayed propagation in content works.

Consider for example if two signing keys A and B are requested, in order.  Imagine that in this case, RPMs are available for both keys, but there's a delay in the content becoming available.

Isn't it possible that you test for existence of A and get False, and then test for existence of B and get True, just because B happened to sync before A or even just because some time passed between your checks on A and B?  So you end up picking B even though you should pick A.  And this is going to happen at random, meaning the overall behavior is going to be slightly flaky.

It sucks, but if we want this to be reliable I think we need to wait the entire timeout while looking only for the *first* file.  Otherwise if you find any file other than the first, you still don't know if the first file is genuinely missing or it's just about to appear.  Basically if we are saying that PUSHSOURCE_SRC_POLL_TIMEOUT is the max possible synchronization delay we expect between storage, then it is not possible to safely conclude that any particular file doesn't exist until after that timeout has been reached.

Alternatively: does Pub actually need to use this with more than one signing_key?  If not, then it would be an option to implement the polling only in the case with a single signing_key so there's only one path to check. Or even to add an option to this source which makes it return the paths without any existence check at all, which also fixes the problem that this polling is going to happen more than once in the case where a file is missing.

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -283,18 +284,38 @@ def _push_items_from_rpm_meta(self, rpm, meta):
...
-            if key:
-                key = key.lower()
-                candidate = os.path.join(build_path, self._pathinfo.signed(meta, key))
-            else:
-                candidate = unsigned_path
-            candidate_paths.append(candidate)
-            if os.path.exists(candidate):
-                rpm_path = candidate
-                rpm_signing_key = key
+        timeout = int(os.getenv("PUSHSOURCE_SRC_POLL_TIMEOUT") or "0")
```
rohanpm, line 287: It'd be preferable if there was less duplicated code between this and SourceWrapper.  For example, could there not be a function with signature:

```
def wait_exists(path: str):
    """Wait until the given path appears to exist, if enabled by environment variables.

    If the file does not exist after a certain timeout, logs a warning.
    """
```

Or if you had to, you could have:

```
def wait_any_exists(path: list[str]) -> Optional[str]:
   """Wait for any of the given paths to exist.

   Returns the first path confirmed to exist, or None if none of the paths
   exist by the deadline.
   """
```

Either one of which would be shareable between SourceWrapper and KojiSource rather than having a similar loop in both places.  But per later comments I doubt the second approach can be done safely anyway.

rohanpm (changes requested):

tests/koji/test_koji.py
```diff
@@ -306,3 +307,103 @@ def test_koji_cache(fake_koji, koji_dir):
 
     # It should have stored something in the cache
     assert cache
+
+
+@patch.dict(
```
rohanpm, line 312: Could you please add the tests in a new file?  We don't need all the koji tests to go into the same file.

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -283,18 +284,38 @@ def _push_items_from_rpm_meta(self, rpm, meta):
...
+                    break
+
+            if rpm_path or i == max_attempts:
+                break
+            LOG.info(
+                "Waiting for %s seconds for any of the following paths to appear: %s",
+                poll_rate,
+                ", ".join(candidate_paths),
+            )
+            time.sleep(poll_rate)
```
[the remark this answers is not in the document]
querti: The code actually raises an error if the path is not present. This is why I didn't add a warning, because the code will have an error message regardless. 

It can be seen at http://10.0.149.4/pub/task/15857/log/stdout.log

src/pushsource/_impl/backend/koji_source.py
querti: The code is changed now so this is no longer a concern, but I did add a condition `if candidate not in candidate_paths:`

src/pushsource/_impl/backend/koji_source.py
querti: Yeah, this did occur to me during implementation, but it seemed like the code didn't care that much about which key will be used in the end.

I changed the logic to poll only for the first key until timeout. Though if the highest prio key fails to appear, a wrong key will be used (instead of push failing, which will happen in other cases of polling).

src/pushsource/_impl/backend/koji_source.py
querti: In the current code iteration the logic seems different enough that it doesn't make sense to add it to the same function. Though the method has become quite long, so if you insist I'll try to think of a way to split the code.

tests/koji/test_koji.py
querti: Sure, done

src/pushsource/_impl/backend/koji_source.py
rohanpm: I don't think they are particularly different.  It seems to me like the SourceWrapper code could be:

```
for item in generator:
  if getattr(item, 'src', None):
    wait_exists(item.src)
  yield item
```

And the code in here could be:

```
if key:
    key = key.lower()
    candidate = os.path.join(build_path, self._pathinfo.signed(meta, key))
    wait_exists(candidate)
```

I suppose you could add an argument to wait_exists to customize the log message, but I'm not sure that's really necessary.

To me that code is much cleaner.  I'm not going to insist on it though, but can you please not put me in that position ("I will improve the code only if you insist on it")?  If you think my suggestion is not an improvement, just say so.

src/pushsource/_impl/backend/koji_source.py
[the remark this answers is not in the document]
rohanpm: Hmm, that error comes from ErrataSource though, which is not the only way KojiSource is used.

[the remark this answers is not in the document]
querti: After separating the polling to its own function, there's now a warning message.

rohanpm (approved):

midnightercz (approved):

PR #259, opened 2022-08-02
Pin attr module

amcmahon-rh:
In a recent update, the attr module removed the "cmp" value from
Attribute classes. However, constructors still expect to receive this
value, it just doesn't get used. That change breaks the fixup_attrs()
function, which tries to copy values to a new Attribute instance. 
Since there's no "cmp" to copy, it doesn't pass the expected value, 
and python has a hissy fit (pun intended) 

This change pins the version of attrs to the previous version. We could
instead hardcode a "cmp" value in.


querti:
I also created a fix: https://github.com/release-engineering/pushsource/pull/260

amcmahon-rh:
> I also created a fix: #260

I like your approach more; TIL of the inspect module :smile: 

 I'll close this PR in favour of yours.

PR #265, opened 2022-09-01
Replaced connection error by Retry error when getting manifest for

midnightercz:
registry source

### DIFF
--- a/src/pushsource/_impl/backend/registry_source.py
+++ b/src/pushsource/_impl/backend/registry_source.py
@@ -109,7 +109,9 @@ class RegistrySource(Source):
                     )
                     self._manifests[source_uri][mtype] = manifest_details
                 except requests.exceptions.RetryError as e:
-                    continue
+                    if "too many 404 error responses" in str(e):
+                        continue
+                    raise e
                 except requests.exceptions.HTTPError as e:
                     if e.response.status_code == 404:
                         continue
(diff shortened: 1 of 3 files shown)


src/pushsource/_impl/backend/registry_source.py
```diff
@@ -108,10 +108,9 @@ def _push_item_from_registry_uri(self, uri, signing_key):
                         manifest_types=[mtype],
                     )
                     self._manifests[source_uri][mtype] = manifest_details
-                except (
-                    requests.exceptions.ConnectionError,
-                    requests.exceptions.HTTPError,
-                ) as e:
+                except requests.exceptions.RetryError as e:
```
rohanpm, line 111: What's the reasoning behind this?  RetryError is what's raised after all retries are exhausted, it's not clear why this should be ignored.

Could you point at a specific problem (i.e. with traceback) which would be solved by this?

midnightercz: Instead of HTTPError, RetryError is raise even on 404. But 404 should be tolerated here.

rohanpm: It looks OK now.  In the version of the code where I posted this comment, you didn't check for 404 but instead caught all kinds of retryerror with no filtering.

test-requirements-py2.txt
```diff
@@ -1,6 +1,6 @@
 jinja2
 pytest
 mock
-rpm-py-installer
+rpm-py-installer < 4.16
```
rohanpm, line 4: If you need this then please do it in its own PR with the reasoning explained.

midnightercz: I accidentally included it in this commit

chandwanitulsi (approved):

PR #283, opened 2022-11-22
Added size attribute to RpmPushItem and FilePushItem

midnightercz:


rohanpm (changes requested):
I don't agree with the concept of this change.  These classes are "dumb" models, basically just a bag of attributes.  I don't think it is in scope for them to have convenience methods added for various ways that you might want to use those attributes.  It opens the door to inconsistency, scope creep and arbitrary design (e.g. why is 'size' a property and not an attribute; why do we have 'size' and not other values returned by stat).  None of the PushItem classes right now have any `@property` and this is by design.

Also, for this particular use-case it seems that it would be better to encourage usage of the standard library.  Python already has a Path object which has all kinds of convenience methods you might want on a path per https://docs.python.org/3/library/pathlib.html#methods .  Like `.exists()`, `.is_file()`, `.read_bytes()` and `.stat().st_size`.  Once we're able to drop python 2 support, it seems a good idea to provide a `Path`-type `src` attribute.  That way all of those convenience methods in the standard library can be accessed easily, rather than having an ad-hoc subset added onto the PushItem classes.

Could you please keep this convenience method only in your project which *uses* the push items?

PR #286, opened 2022-12-05
Only poll for existence if item source is a file

querti:
Recently, a polling mechanism was introducted to pushsource to check if a file exists before continueing the push. This was necessary because mounted volumes (where the files are located) are snapmirrored and there's a time delay before new files are propagated to the snapmirror volumes. There is an unwanted side effect where all push items are checked, even those that don't contain a file in their source (namely container images that contain the image URL instead). This is undesirable behavior, because image URL is not a file and thus will never be found on disk. This bug significantly slows the pushes, since the items will be polled until timeout is reached. Fix it by only checking push item sources that start with "/", indicating that they are local files.

Refers to CLOUDDST-16707

### DIFF
--- a/src/pushsource/_impl/source.py
+++ b/src/pushsource/_impl/source.py
@@ -77,7 +77,11 @@ class SourceWrapper(object):
 
         generator = self.__delegate.__iter__()
         for item in generator:
-            if not hasattr(item, "src") or not item.src:
+            if (
+                not hasattr(item, "src")
+                or not item.src
+                or not item.src.startswith("/")
+            ):
                 yield item
             else:
                 wait_exist(item.src, timeout, poll_rate)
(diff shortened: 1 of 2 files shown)


src/pushsource/_impl/source.py
```diff
@@ -77,7 +77,11 @@ def __iter__(self):
 
         generator = self.__delegate.__iter__()
         for item in generator:
-            if not hasattr(item, "src") or not item.src:
+            if (
+                not hasattr(item, "src")
+                or not item.src
+                or not item.src.startswith("/mnt")
```
midnightercz, line 83: it would be nice to have this configurable or maybe check just for  /

midnightercz (changes requested):

src/pushsource/_impl/source.py
querti: Changed to /

rohanpm (approved):

emilyzheng (approved):

midnightercz (approved):

PR #291, opened 2022-12-19
Add Azure VHD & Virtual Machine Images Push Items

JAVGan:
This commit implement the following changes:

  - Create the `VHDPushItem` model to represent Azure's VHD images

  - Create a new `PushItem` model named `VMIPush` item to be the parent class for `AmiPushItem`

  - Created the model `VMIRelease` to replace `AmiRelease` for all marketplaces as the common data are applied for both Azure and AWS. To avoid breaking this library's userspace the AmiRelease now inherit rom VMIRelease but doesn't add any new field.

  - Modify some attributes from `AmiPushItem` to be optional in order to allow loading from `KojiSource` in the future while retrieving the remaining data from an external service.

The `VMIPushItem` and `VMIRelease` now represents the common interface for all VM images which should go to a proper cloud marketplace.

The `VHDPushItem` will be used in future to be handled from `KojiSource` along with `AmiPushItem`.

Refers to EXDSP-1777


JAVGan:
@Tojaj @lslebodn @rohanpm PTAL whenever possible

JAVGan:
@Tojaj @lslebodn @rohanpm PTAL whenever possible

src/pushsource/_impl/model/ami.py
```diff
@@ -1,46 +1,20 @@
...
-    base_version = attr.ib(type=str, default=None, validator=optional_str)
-    """For layered products, version of the base product for which the image should be used."""
+class AmiRelease(VMIRelease):
+    """
+    Release metadata associated with an AMI.
 
-    variant = attr.ib(type=str, default=None, validator=optional_str)
-    """Variant of this image's product (only for products which have variants).  For example,
-    "Server", for Red Hat Enterprise Linux Server."""
+    This class has the same attributes ass VMIRelease and its only kept
```
rohanpm, line 12:
```suggestion
    This class has the same attributes as VMIRelease and it's only kept
```

src/pushsource/_impl/model/ami.py
```diff
@@ -1,46 +1,20 @@
...
+    for external tools compatibility.
```
rohanpm, line 13: Just say "backwards compatibility", it is enough justification to not remove something from an API.


```suggestion
    for backwards compatibility.
```

src/pushsource/_impl/model/ami.py
```diff
@@ -1,46 +1,20 @@
...
+    """
 
-    type = attr.ib(type=str, default=None, validator=optional_str)
-    """Release type, typically "ga" or "beta"."""
+    def __init__(self, *args, **kwargs):
```
rohanpm, line 16: Hmm, is this needed actually?  Shouldn't this just work without having to explicitly declare an `__init__` ?

src/pushsource/_impl/model/azure.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+
+    generation = attr.ib(type=str, default="V2", validator=optional(in_(VHDGeneration)))
+    """VHD generation. It can be ``V1`` or ``V2``."""
+
+    sku_id = attr.ib(type=str, default=None, validator=optional_str)
+    """SKU ID for the VM image generation, generally being the Plan ID."""
+
+    support_legacy = attr.ib(type=bool, default=False, validator=instance_of(bool))
+    """If the image ``generation == V2`` but also support ``V1``."""
```
rohanpm, line 31:
```suggestion
    """If the image ``generation == V2`` but also supports ``V1``."""
```

src/pushsource/_impl/model/azure.py
```diff
@@ -0,0 +1,55 @@
...
+
+    legacy_sku_id = attr.ib(
+        type=str, default=None, validator=optional(in_(VHDGeneration))
+    )
+    """If the ``V2`` image also support ``V1`` what is the ``V1`` SKU ID."""
```
rohanpm, line 36:
```suggestion
    """The ``V1`` SKU ID (only if ``support_legacy`` is true)."""
```

src/pushsource/__init__.py
```diff
@@ -17,9 +17,12 @@
     ContainerImagePullInfo,
     SourceContainerImagePushItem,
     OperatorManifestPushItem,
+    VMIPushItem,
```
rohanpm, line 20: There is a mismatch in naming scheme here, e.g. VMIPushItem vs AmiPushItem.

I was going to recommend that your classes should be renamed to be consistent with those present, e.g. VmiPushItem.  However, PEP8 says we should "capitalize all the letters of the acronym" so I guess it's the existing classes which are wrong.  Let's stick with the PEP8-consistent names and fix up the others later.

docs/model/azure.rst
```diff
@@ -0,0 +1,5 @@
+Push items: VHD
```
rohanpm, line 1: A change will be needed to docs/index.rst as well in order to make these pages reachable.

src/pushsource/_impl/model/azure.py
```diff
@@ -0,0 +1,55 @@
...
+    recommended_sizes = attr.ib(
+        type=list, default=attr.Factory(frozenlist), converter=frozenlist
+    )
+    """List of recommended VM sizes in Azure, for example ['a1-standard', 'a5', 'a7']"""
+
+    sas_uri = attr.ib(type=str, default=None, validator=optional_str)
+    """The SAS URI to be associated with the disk_version."""
+
+    @disk_version.validator
+    def _validate_regex(self, attribute, value):
```
rohanpm, line 50: Since there is a class hierarchy here, I think this should be named with a double leading underscore `(def __validate_whatever)` to activate name mangling.  Otherwise, if someone ever introduced a method of the same name in a subclass or superclass of this, then one method is going to shadow another.

Also, though less important, `validate_regex` isn't a great name seeing as there could later be more than one attribute on the class which needs to be validated by a regex.  Naming it after the attribute being validated seems to make more sense.

src/pushsource/_impl/model/vms.py
```diff
@@ -0,0 +1,57 @@
...
+
+@attr.s()
+class VMIPushItem(PushItem):
+    """A :class:`~pushsource.PushItem` representing a generic VM image.
+
+    This class is used as a parent class for different types of VM images.
+    """
+
+    release = attr.ib(
+        type=VMIRelease, default=None, validator=optional(instance_of(VMIRelease))
```
rohanpm, line 52: I think the validator here might need to be more complicated.  Currently, it will allow mixing incompatible VMIRelease subclasses and VMIPushItem subclasses.  For example you'll be able to put a generic VMIRelease onto an AmiPushItem, or an AmiRelease onto a VHDPushItem.

I don't think that's a good idea.  Given that, prior to now, AmiPushItem.release has always been an AmiRelease, it would have been fair for all code using it to assume that would be the case, and arguably a backwards-incompatible change if we start allowing objects where this is not the case.  Easier to reason about if the release on a push item is guaranteed to always be of the correct type associated with that item.

Would it work having a class-level private attribute which can be overridden by subclasses?  e.g.

```python
class VMIPushItem(PushItem):
    _RELEASE_TYPE = VMIRelease

    ...
    @release.validator()
    def __validate_release(self, attribute, value):
        ... # validate that value is either None or isinstance(self._RELEASE_TYPE)
```

And then a subclass can ensure that the correct type of release is always used by:

```python
class AmiPushItem(VMIPushItem):
    _RELEASE_TYPE = AmiRelease
```

rohanpm (changes requested):

src/pushsource/_impl/model/ami.py
JAVGan: Yes, it works. I'll adjust it, thanks!

src/pushsource/_impl/model/vms.py
JAVGan: Yes, I like this idea!

I'll change it, thanks!

docs/model/azure.rst
JAVGan: Fixed in `index.rst`, PTAL

src/pushsource/_impl/model/vms.py
JAVGan: I've fixed it, PTAL again

JAVGan:
@rohanpm I've added your suggestions and fixed the points you mentioned except the class names as the acronym follows PEP8.

PTAL again

src/pushsource/_impl/model/ami.py
JAVGan: Fixed

src/pushsource/_impl/model/ami.py
JAVGan: Fixed

src/pushsource/_impl/model/ami.py
JAVGan: Fixed

src/pushsource/_impl/model/azure.py
JAVGan: Fixed

src/pushsource/_impl/model/azure.py
JAVGan: Fixed

rohanpm (approved):

chandwanitulsi (approved):

src/pushsource/_impl/model/azure.py
```diff
@@ -0,0 +1,55 @@
...
+    """A :class:`~pushsource.PushItem` representing a Virtual Hard Disk (or VHD).
+
+    For push items of this type, the :meth:`~pushsource.PushItem.src` attribute
+    refers to a file containing the VHD file..
+
+    This library does not verify that the referenced file is a valid
+    snapshot.
+    """
+
+    generation = attr.ib(type=str, default="V2", validator=optional(in_(VHDGeneration)))
```
lslebodn, line 24: @JAVGan woudl you mind to explain why there is "optional"  validator?
IMHO it should be required in case of default value.
But maybe I do not understand something :-)

src/pushsource/_impl/model/azure.py
```diff
@@ -0,0 +1,55 @@
...
+    disk_version = attr.ib(type=str, default=None)
+    """The image's disk version according to Azure's format {int}.{int}.{int}"""
+
+    recommended_sizes = attr.ib(
+        type=list, default=attr.Factory(frozenlist), converter=frozenlist
+    )
+    """List of recommended VM sizes in Azure, for example ['a1-standard', 'a5', 'a7']"""
+
+    sas_uri = attr.ib(type=str, default=None, validator=optional_str)
+    """The SAS URI to be associated with the disk_version."""
```
lslebodn, line 47: I was under impression that SAS URI will be required.
Maybe it would be enough to to add a comment when value can be omitted from the object.

lslebodn (commented):
Few tiny comments.

And one more generic question/comment. Sorry for my ignorance.

> The VHDPushItem will be used in future to be handled from KojiSource along with AmiPushItem.

I assume changes related to KojiSource will be done in different PR.
Or there is something hidden in the code?

src/pushsource/_impl/model/azure.py
JAVGan: I've set it as optional so the object can be created without passing the `generation` as argument and it should default to "V2".

Without this "optional" it would be required to pass the argument `generation` during the instantiation and we can't retrieve this value from koji directly

src/pushsource/_impl/model/azure.py
JAVGan: It will be computed _after_, it is, not during the source from Koji but when the pubtool finishes the upload or retrieves an already uploaded object

JAVGan:
> Few tiny comments.
> 
> And one more generic question/comment. Sorry for my ignorance.
> 
> > The VHDPushItem will be used in future to be handled from KojiSource along with AmiPushItem.
> 
> I assume changes related to KojiSource will be done in different PR. Or there is something hidden in the code?

It will be added in a next PR.

This one just create the models and I'm already working in a next one to source from Koji but I prefer to have this merged first.

src/pushsource/_impl/model/azure.py
lslebodn: > Without this "optional" it would be required to pass the argument generation during the instantiation and we can't retrieve this value from koji directly

would you mind to add such doc comment?

src/pushsource/_impl/model/azure.py
lslebodn: OK; would you mind to add similar comment to the code?

src/pushsource/_impl/model/azure.py
JAVGan: I've added in the class description, PTAL again

src/pushsource/_impl/model/azure.py
JAVGan: I've added in the class description, PTAL again

lslebodn: tyvm

src/pushsource/_impl/model/azure.py
lslebodn: tyvm

lslebodn (approved):
LGTM

JAVGan:
@rohanpm can you please help us to merge this?

It's already approved by @lslebodn which just asked to add a note in the comments.

Regards

src/pushsource/_impl/model/azure.py
rohanpm: > I've set it as optional so the object can be created without passing the `generation` as argument and it should default to "V2".

This doesn't sound right (as a justification for 'optional'), because the above is already achieved by setting default="V2".

Including optional in the validator has the *added* effect that it is possible to create an object having `generation=None`, but this is not "without passing the generation as argument", this is explicitly passing an argument of value None.  So, do you want that to be possible/do you expect to do that or not?

If not sure then I'd suggest removing it (not allowing None), as it's always easier to widen the range of allowed values later, than it is to narrow them (which is backwards-incompatible).

src/pushsource/_impl/model/azure.py
```diff
@@ -0,0 +1,60 @@
...
+
+    legacy_sku_id = attr.ib(
+        type=str, default=None, validator=optional(in_(VHDGeneration))
+    )
+    """The ``V1`` SKU ID (only if ``support_legacy`` is true)."""
+
+    disk_version = attr.ib(type=str, default=None)
+    """The image's disk version according to Azure's format {int}.{int}.{int}"""
+
+    recommended_sizes = attr.ib(
```
rohanpm, line 46: Does order matter for the elements in this list?  For example is the first size "more recommended" than the next?

If order doesn't matter, you might consider making the converter sort() them so that there aren't multiple possible representations of the "same" value.

JAVGan: It doesn't matter. Sure, I'll add the converter, thanks!

JAVGan: @rohanpm I've added a sorting for it, please let me know if it's good now

JAVGan:
@rohanpm please let me know if this one looks good.. If it is please merge as well

PR #294, opened 2023-01-04
EXDSP-1781: Allow KojiSource to retrieve VMIPushItems

JAVGan:
This PR allows `pushsource` to load Virtual Machine Images through `KojiSource`.

The VM images will come as a `VMIPushItem` or as its subclasses: `AmiPushItem` / `VHDPushItem`.

To load the VM images the application may use:

```{python}
from pushsource import Source, VMIPushItem

with Source.get('koji:https://koji.fedoraproject.org/kojihub?vmi_build=img1,img2,...') as source:
    for item in source:
        ...  # do something
```

**NOTE:** As Koji doesn't contain all metadata for `VMIPushItem` and its subclasses it only load the common values which can be found on it. Missing information as well as specific data required for Azure or AWS may be fetched from an external service.

Refers to EXDSP-1781

### DIFF
--- a/docs/sources/koji.rst
+++ b/docs/sources/koji.rst
@@ -11,6 +11,7 @@ Supported content types:
 * RPMs
 * container images (with operator manifests if applicable)
 * modulemd YAML streams (yum repo metadata)
+* virtual machine images
 
 Note that many features of the koji source requires having koji's volume(s) mounted
 locally.
@@ -77,6 +78,21 @@ Push items produced in this case may include:
 - :class:`~pushsource.OperatorManifestPushItem` - if images have attached operator
   manifest archives
 
+Accessing virtual machine images
+...............................
+
+Use the ``vmi_build`` parameter to request virtual machine images from one or
+more builds. Builds can be specified by NVR. Builds should be produced by
+`Image Builder`_ or should use compatible metadata.
+
+``koji:https://koji.fedoraproject.org/kojihub?vmi_build=fedora-azure-37-1.7``
+
+Push items produced in this case may include:
+
+- :class:`~pushsource.VMIPushItem` - one for each available image
+- :class:`~pushsource.VHDPushItem` - for Azure VHD images.
+- :class:`~pushsource.AmiPushItem` - for AMI raw images.
+
 Setting the destination for push items
 ......................................
 
@@ -110,3 +126,4 @@ Python API reference
 
 
 .. _OSBS: https://osbs.readthedocs.io/
+.. _Image Builder: https://www.osbuild.org/guides/image-builder-service/image-builder-koji.html
--- a/src/pushsource/_impl/backend/koji_source.py
+++ b/src/pushsource/_impl/backend/koji_source.py
@@ -19,6 +19,9 @@ from ..model import (
     OperatorManifestPushItem,
     ContainerImagePushItem,
     SourceContainerImagePushItem,
+    VMIPushItem,
+    AmiPushItem,
+    VHDPushItem,
 )
 from ..helpers import (
     list_argument,
@@ -122,6 +125,7 @@ class KojiSource(Source):
         module_build=None,
         module_filter_filename=None,
         container_build=None,
+        vmi_build=None,
         signing_key=None,
         basedir=None,
         threads=4,
@@ -167,6 +171,11 @@ class KojiSource(Source):
                 build(s), along with associated artifacts such as operator manifest
                 bundles.
 
+            vmi_build (list[str, int])
+                Virtual Machine Image identifier(s). Can be koji build IDs (integers)
+                or build NVRs.
+                The source will yield all VMIs identified by this list.
+
             signing_key (list[str])
                 GPG signing key ID(s). If provided, content must be signed
                 using one of the provided keys. Include ``None`` if unsigned
@@ -204,6 +213,7 @@ class KojiSource(Source):
             module_filter_filename, retain_none=True
         )
         self._container_build = [try_int(x) for x in list_argument(container_build)]
+        self._vmi_build = [try_int(x) for x in list_argument(vmi_build)]
         self._signing_key = list_argument(signing_key)
         self._dest = list_argument(dest)
         self._timeout = timeout
@@ -504,6 +514,110 @@ class KojiSource(Source):
 
         return out
 
+    def _push_items_from_vmi_build(self, nvr, meta):
+        LOG.debug("Looking for virtual machine images on %s, %s", nvr, meta)
+
+        if not meta:
+            message = "Virtual machine image build not found in koji: %s" % nvr
+            LOG.debug(message)
+            raise ValueError(message)
+
+        # VMI has the {'typeinfo': {'image':{}}} on extra
+        extra = meta.get("extra") or {}
+        typeinfo = extra.get("typeinfo") or {}
+        image = typeinfo.get("image")
+
+        # Container images have the key `container_koji_task_id` set on extras:
+        # https://github.com/containerbuildsystem/atomic-reactor/blob/master/docs/koji.md?plain=1#L79
+        if extra.get("container_koji_task_id"):
+            message = "The build %s is for container images." % nvr
+            LOG.debug(message)
+            raise ValueError(message)
+
+        if image is None:
+            message = "Build %s not recognized as a virtual machine image build" % nvr
+            LOG.debug(message)
+            raise ValueError(message)
+
+        build_id = meta["id"]
+
+        archives = self._get_archives(build_id)
+
+        # Supported types of VMI archives
+        ami_types = ["raw", "raw-xz"]
+        azure_types = ["vhd", "vhd-compressed"]
+        cloud_types = {}
+
+        for k in ami_types:
+            cloud_types.setdefault(k, AmiPushItem)
+
+        for k in azure_types:
+            cloud_types.setdefault(k, VHDPushItem)
+
+        # Collect all known VM image archives
+        vmi_archives = [
+            elem
+            for elem in archives
+            if elem.get("btype") == "image"
+            and elem.get("type_name") in cloud_types.keys()
+        ]
+
+        out = []
+
+        # Generate the respective PushItem type for each archive
+        for archive in vmi_archives:
+            path = self._pathinfo.typedir(meta, archive["btype"])
+            item_src = os.path.join(path, archive["filename"])
+
+            extra = archive.get("extra") or {}
+            image = extra.get("image") or {}
+            arch = image.get("arch")
+
+            klass = cloud_types.get(archive["type_name"]) or VMIPushItem
+
+            rel_klass = klass._RELEASE_TYPE
+
+            # The given name will be composed by the product name and cloud
+            # marketpalce name, like "rhel-azure" or "rhel-sap-azure".
+            # We just want to get the product name without the cloud name,
+            # like "rhel" and "rhel-sap" in our examples respectively.
+            splitted_name = meta["name"].split("-")
+            if len(splitted_name) > 1:
+                product = "-".join(splitted_name[:-1])
+            else:
+                product = splitted_name[0]
+
+            # Try to get the resping version from release, if available
+            respin = try_int(meta["release"])
+            respin = 0 if not isinstance(respin, int) else respin
+
+            release = rel_klass(
+                arch=arch,
+                date=(meta.get("completion_time") or "").split(" ")[0],
+                product=product.upper(),
+                version=meta["version"],
+                respin=respin,
+            )
+
+            out.append(
+                klass(
+                    name=archive["filename"],
+                    description="",
+                    dest=self._dest,
+                    src=item_src,
+                    build=nvr,
+                    build_info=KojiBuildInfo(
+                        name=meta["name"],
+                        version=meta["version"],
+                        release=meta["release"],
+                    ),
+                    md5sum=archive.get("checksum"),
+                    release=release,
+                )
+            )
+
+        return out
+
     def _get_operator_item(self, nvr, meta, archives):
         extra = meta.get("extra") or {}
         typeinfo = extra.get("typeinfo") or {}
@@ -570,6 +684,19 @@ class KojiSource(Source):
         return [
             f_map(f, partial(self._push_items_from_container_build, build))
             for f, build in build_fs
+        ]
+
+    def _vmi_futures(self):
+        # Get info from each requested virtual machine image build.
+        vmi_build_fs = [
+            (self._executor.submit(self._get_build, build), build)
+            for build in self._vmi_build
+        ]
+
+        # Convert them to lists of push items
+        return [
+            f_map(f, partial(self._push_items_from_vmi_build, build))
+            for f, build in vmi_build_fs
         ]
 
     def _do_fetch(self, koji_queue, exceptions):
@@ -649,6 +776,10 @@ class KojiSource(Source):
         for build_id in self._container_build:
             koji_queue.put(GetBuildCommand(ident=build_id, list_archives=True))
 
+        # We'll need to obtain all virtual machine image builds
+        for build_id in self._vmi_build:
+            koji_queue.put(GetBuildCommand(ident=build_id, list_archives=True))
+
         # Put some threads to work on the queue.
         fetch_exceptions = []
         fetch_threads = [
@@ -692,7 +823,10 @@ class KojiSource(Source):
         assert koji_queue.empty()
 
         push_items_fs = (
-            self._modulemd_futures() + self._rpm_futures() + self._container_futures()
+            self._modulemd_futures()
+            + self._rpm_futures()
+            + self._container_futures()
+            + self._vmi_futures()
         )
 
         completed_fs = as_completed_with_timeout_reset(
--- a/tests/koji/test_koji_vmi.py
+++ b/tests/koji/test_koji_vmi.py
@@ -1,0 +1,228 @@
+import os
+
+from pytest import raises
+
+from pushsource import (
+    Source,
+    AmiPushItem,
+    VHDPushItem,
+    VMIPushItem,
+    AmiRelease,
+    VMIRelease,
+)
+
+
+def test_koji_vmi_notfound(fake_koji):
+    """koji source referencing nonexistent VMI build will fail."""
+
+    source = Source.get("koji:https://koji.example.com/?vmi_build=notexist-1.2.3")
+    with raises(ValueError) as exc_info:
+        list(source)
+
+    assert "Virtual machine image build not found in koji: notexist-1.2.3" in str(
+        exc_info.value
+    )
+
+
+def test_koji_not_vmi_invalid(fake_koji, koji_dir):
+    """Koji source referencing an invalid build will fail."""
+
+    source = Source.get(
+        "koji:https://koji.example.com/?vmi_build=foobuild-1.0-1", basedir=koji_dir
+    )
+
+    fake_koji.build_data[1234] = {
+        "id": 1234,
+        "name": "foobuild",
+        "version": "1.0",
+        "release": "1",
+        "nvr": "foobuild-1.0-1",
+        "completion_time": "2022-12-20 12:30:25",
+    }
+    fake_koji.build_data["foobuild-1.0-1"] = fake_koji.build_data[1234]
+
+    with raises(ValueError) as exc_info:
+        list(source)
+
+    assert (
+        "Build foobuild-1.0-1 not recognized as a virtual machine image build"
+        in str(exc_info.value)
+    )
+
+
+def test_koji_not_vmi_container(fake_koji, koji_dir):
+    """Koji source referencing a container will fail."""
+
+    source = Source.get(
+        "koji:https://koji.example.com/?vmi_build=foobuild-1.0-1", basedir=koji_dir
+    )
+
+    # Container image build
+    fake_koji.build_data[1234] = {
+        "id": 1234,
+        "name": "foobuild",
+        "version": "1.0",
+        "release": "1",
+        "nvr": "foobuild-1.0-1",
+        "completion_time": "2022-12-20 12:30:25",
+        "extra": {"typeinfo": {"image": {}}, "container_koji_task_id": "container"},
+    }
+    fake_koji.build_data["foobuild-1.0-1"] = fake_koji.build_data[1234]
+    with raises(ValueError) as exc_info:
+        list(source)
+
+    assert "The build foobuild-1.0-1 is for container images." in str(exc_info.value)
+
+
+def test_koji_vmis(fake_koji, koji_dir):
+    """Koji source yields requested Virtual Machine Images."""
+
+    name = "foobuild"
+    version = "1.0"
+    release = "20230110.sp.22"
+    nvr = "%s-%s-%s" % (name, version, release)
+
+    source = Source.get(
+        "koji:https://koji.example.com/?vmi_build=%s" % nvr, basedir=koji_dir
+    )
+
+    # It should not have done anything yet (lazy loading)
+    assert not fake_koji.last_url
+
+    # Insert some data
+    fake_koji.build_data[1234] = {
+        "id": 1234,
+        "name": name,
+        "version": version,
+        "release": release,
+        "nvr": nvr,
+        "completion_time": "2022-12-20 12:30:25",
+        "extra": {"typeinfo": {"image": {}}},
+    }
+    fake_koji.build_data[nvr] = fake_koji.build_data[1234]
+
+    imgs = {
+        "foo-1.0-20230110.sp.22.img": "generic",
+        "foo-1.0-20230110.sp.22.raw.xz": "raw-xz",
+        "foo-1.0-20230110.sp.22.vhd.xz": "vhd-compressed",
+    }
+
+    archives = []
+    for fname, tname in imgs.items():
+        archives.append(
+            {
+                "btype": "image",
+                "filename": fname,
+                "checksum": "00000000000000000000000000000000",
+                "extra": {"image": {"arch": "x86_64"}},
+                "type_name": tname,
+                "build_id": 1234,
+                "nvr": nvr,
+            }
+        )
+
+    fake_koji.insert_archives(archives=archives, build_nvr=nvr)
+
+    # Eagerly fetch
+    items = list(source)
+
+    # It should have returned only valid push items for the modules
+    assert len(items) == 2
+
+    items = sorted(items, key=lambda pi: pi.name)
+
+    klasses = [AmiPushItem, VHDPushItem]
+    index = 0
+    imgs.pop("foo-1.0-20230110.sp.22.img")  # Invalid image filtered out by KojiSource
+
+    for klass in klasses:
+        fname = sorted(list(imgs.keys()))[index]
+
+        rel_klass = AmiRelease if klass == AmiPushItem else VMIRelease
+
+        rel_obj = rel_klass(
+            arch="x86_64",
+            date="2022-12-20",
+            product="FOOBUILD",
+            version=version,
+            respin=0,
+        )
+
+        assert items[index] == klass(
+            name=fname,
+            state="PENDING",
+            src=os.path.join(
+                koji_dir, "packages/foobuild/1.0/20230110.sp.22/images/%s" % fname
+            ),
+            dest=[],
+            description="",
+            md5sum="00000000000000000000000000000000",
+            sha256sum=None,
+            origin=None,
+            build=nvr,
+            signing_key=None,
+            release=rel_obj,
+        )
+        index += 1
+
+
+def test_koji_vmi_compound_product_name(fake_koji, koji_dir):
+    source = Source.get(
+        "koji:https://koji.example.com/?vmi_build=foobuild-azure-1.0-1",
+        basedir=koji_dir,
+    )
+
+    # Insert some data
+    fake_koji.build_data[1234] = {
+        "id": 1234,
+        "name": "foobuild-azure",
+        "version": "1.0",
+        "release": "1",
+        "nvr": "foobuild-azure-1.0-1",
+        "completion_time": "2022-12-20 12:30:25",
+        "extra": {"typeinfo": {"image": {}}},
+    }
+    fake_koji.build_data["foobuild-azure-1.0-1"] = fake_koji.build_data[1234]
+
+    fake_koji.insert_archives(
+        archives=[
+            {
+                "btype": "image",
+                "filename": "foo-azure-1.0-1.raw.xz",
+                "checksum": "00000000000000000000000000000000",
+                "extra": {"image": {"arch": "x86_64"}},
+                "type_name": "raw-xz",
+                "build_id": 1234,
+                "nvr": "foobuild-azure-1.0-1",
+            }
+        ],
+        build_nvr="foobuild-azure-1.0-1",
+    )
+
+    # Eagerly fetch
+    items = list(source)
+
+    # For product the name "foobuild-azure" should become only "foobuild" in uppercase
+    rel_obj = AmiRelease(
+        arch="x86_64",
+        date="2022-12-20",
+        product="FOOBUILD",
+        version="1.0",
+        respin=1,
+    )
+
+    assert items[0] == AmiPushItem(
+        name="foo-azure-1.0-1.raw.xz",
+        state="PENDING",
+        src=os.path.join(
+            koji_dir, "packages/foobuild-azure/1.0/1/images/foo-azure-1.0-1.raw.xz"
+        ),
+        dest=[],
+        description="",
+        md5sum="00000000000000000000000000000000",
+        sha256sum=None,
+        origin=None,
+        build="foobuild-azure-1.0-1",
+        signing_key=None,
+        release=rel_obj,
+    )


src/pushsource/_impl/backend/koji_source.py
```diff
@@ -509,6 +521,127 @@ def _push_items_from_container_build(self, nvr, meta):
...
+                #   - billing_codes
+                #
+                # VHDPushItem:
+                #   - generation
+                #   - sku_id
+                #   - support_legacy
+                #   - legacy_sku_id
+                #   - disk_version
+                #   - recommended_sizes
+                #   - sas_uri
```
JAVGan, line 626: @Tojaj @lslebodn these are the missing metadata from Koji. 

For `AmiPushItem` they're available from Staged Source but here we can't find them.

For `VHDPushItem` we also need everything but `sas_uri` which will be generated after the upload.

JAVGan: We have 2 options here:

1. Ask osbuild team to provide the missing metadata depending on the type of image (VHD / RAW)
2. Include the missing pieces in StArMap but it will require the declaration on  `meta` for all affected policies.

Particularly, if possible I would like to have the option 1 in place as our source of information must be Koji and not another service... WDYT?

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -509,6 +521,127 @@ def _push_items_from_container_build(self, nvr, meta):
...
+            cloud_types.setdefault(k, VHDPushItem)
+
+        # Collect all image archives which are not containers (tar)
+        # FIXME: Need to verify whether it's better to limit only to
+        # known extensions (like VHD, AMI) or not
+        vmi_archives = [
+            elem
+            for elem in archives
+            if elem.get("btype") == "image" and elem.get("type_name") != "tar"
+        ]
```
JAVGan, line 564: @Tojaj @lslebodn @jajreidy in your opinion does it make sense for `KojiSource` to filter the image types only for VHD / AMI  or should it bring all image archives (which are not containers) whatever type it is?

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -509,6 +521,118 @@ def _push_items_from_container_build(self, nvr, meta):
...
+                #   - recommended_sizes
+                #   - sas_uri
+                klass(
+                    name=archive["filename"],
+                    description="",
+                    dest=self._dest,
+                    src=item_src,
+                    build=nvr,
+                    build_info=KojiBuildInfo(
+                        name=meta["name"],
```
JAVGan, line 625: @Tojaj @lslebodn @jajreidy this is the updated list of missing metadata.

For `Azure` the at least the following metadata must be present before publishing:

- disk_version: should come from somewhere: OSBS or StArMap.
- sas_uri (will come from upload)

The other ones have the following defaults:
- generation: Defaults to `V2`
- sku_id: Defaults to `{plan_name}`
- support_legacy: Defaults to `False`
- legacy_sku_id: Defaults to `{plan_name}-gen1`
- recommended_sizes: Defaults to `[]`

@jajreidy do you have the list of required metadata for AWS as well? Maybe we can update the doc with all this information, WDYT?

src/pushsource/_impl/backend/koji_source.py
jajreidy: Do you know what image type that's going to be used for GCP? I know for amazon AMI can be a few file types.

JAVGan: No idea... But yes, the disadvantage of filtering here is that we'll need to always keep the filter up to date...

JAVGan:
@rohanpm @lslebodn @Tojaj @jajreidy please review.

This PR allows `KojiSource` to load the models implemented on !291 along with `AmiPushSource` (which was just available trough `staged`)

docs/sources/koji.rst
```diff
@@ -77,6 +78,21 @@ Push items produced in this case may include:
 - :class:`~pushsource.OperatorManifestPushItem` - if images have attached operator
   manifest archives
 
+Acessing virtual machine images
+...............................
+
+Use the ``vmi_build`` parameter to request virtual machine images from one or
+more builds. Builds can be specified by NVR. Builds should be produced by
+`OSBS`_ or should use compatible metadata.
```
rohanpm, line 86: > Builds should be produced by OSBS or should use compatible metadata.

Is this true?  OSBS is producing these builds?

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -509,6 +521,94 @@ def _push_items_from_container_build(self, nvr, meta):
...
+            path = self._pathinfo.typedir(meta, archive["btype"])
+            item_src = os.path.join(path, archive["filename"])
+
+            extra = archive.get("extra") or {}
+            image = extra.get("image") or {}
+            arch = image.get("arch")
+
+            klass = cloud_types.get(archive["type_name"]) or VMIPushItem
+
+            rel_klass = AmiRelease if klass == AmiPushItem else VMIRelease
```
rohanpm, line 577: Minor: you could consider doing this instead:

```suggestion
            rel_klass = klass._RELEASE_TYPE
```

Then you don't need to update this code if cloud_types later supports more types with their own release class.

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -509,6 +521,94 @@ def _push_items_from_container_build(self, nvr, meta):
...
+
+            splitted_name = meta["name"].split("-")
+            if len(splitted_name) > 1:
+                product = "-".join(splitted_name[:-1])
+            else:
+                product = splitted_name[0]
+
+            release = rel_klass(
+                arch=arch,
+                date=meta.get("completion_time", "").split(" ")[0],
```
rohanpm, line 587: Minor: if you want this field to default to an empty string, it's probably better to be more defensive like this:

```suggestion
                date=(meta.get("completion_time") or "").split(" ")[0],
```

Right now, with the way you've written it, if completion_time is present but null then it'll crash.

docs/sources/koji.rst
```diff
@@ -77,6 +78,21 @@ Push items produced in this case may include:
...
+
+``koji:https://koji.fedoraproject.org/kojihub?vmi_build=fedora-azure-37-1.7.x86_64.vhd.xz``
```
rohanpm, line 88: Does this actually exist in fedora koji?  If so, can you please give me a link, as I can't find it.  Or if it doesn't exist, can you please give me a link to a representative build in any koji?

The main reason I'm asking is because this looks like a filename, not a build NVR, while all the docs say it should be a build NVR.  I want to see a real build so I can understand what one of these builds looks like.

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -509,6 +521,94 @@ def _push_items_from_container_build(self, nvr, meta):
...
+
+            extra = archive.get("extra") or {}
+            image = extra.get("image") or {}
+            arch = image.get("arch")
+
+            klass = cloud_types.get(archive["type_name"]) or VMIPushItem
+
+            rel_klass = AmiRelease if klass == AmiPushItem else VMIRelease
+
+            splitted_name = meta["name"].split("-")
```
rohanpm, line 579: It's hard to understand if this is correct, could you please put a brief comment before this line with an example value or two of `meta["name"]` ?

tests/koji/test_koji_vmi.py
```diff
@@ -0,0 +1,204 @@
...
+        "Build foobuild-1.0-1 not recognized as a virtual machine image build"
+        in str(exc_info.value)
+    )
+
+
+def test_koji_vmis(fake_koji, koji_dir):
+    """Koji source yields requested Virtual Machine Images."""
+
+    nvr = "foobuild-1.0-1"
+    r, v, n = nvr[-1::-1].split("-", 2)
```
rohanpm, line 57: I think it would be better to just write out the assignments of `name`, `version`, `release` in full...

There is after all no need for this code to support arbitrary NVRs since the only handled value is hardcoded anyway.  It would be easier to read if it simply said name="foobuild", version="1.0", release="1" and would reduce the possibility of the test itself being buggy.

tests/koji/test_koji_vmi.py
```diff
@@ -0,0 +1,204 @@
...
+                "extra": {"image": {"arch": "x86_64"}},
+                "type_name": tname,
+                "build_id": 1234,
+                "nvr": nvr,
+            }
+        )
+
+    fake_koji.insert_archives(archives=archives, build_nvr=nvr)
+
+    # raise Exception(fake_koji.build_data)
```
rohanpm, line 103: Could you clean up this line please?

tests/koji/test_koji_vmi.py
```diff
@@ -0,0 +1,204 @@
...
+        fname = sorted(list(imgs.keys()))[index]
+
+        rel_klass = AmiRelease if klass == AmiPushItem else VMIRelease
+
+        rel_obj = rel_klass(
+            arch="x86_64",
+            date="2022-12-20",
+            product=name.upper(),
+            version=version,
+            respin=int(r[-1::-1]),
```
rohanpm, line 126: Again, I think it is better to just hardcode the value for comparison so it can be seen at a glance whether or not the value makes sense.

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -509,6 +521,94 @@ def _push_items_from_container_build(self, nvr, meta):
...
+            if len(splitted_name) > 1:
+                product = "-".join(splitted_name[:-1])
+            else:
+                product = splitted_name[0]
+
+            release = rel_klass(
+                arch=arch,
+                date=meta.get("completion_time", "").split(" ")[0],
+                product=product.upper(),
```
rohanpm, line 588: Where does the requirement to be uppercase come from?

The main reason I ask is that it doesn't seem like this would be koji-specific.  Maybe it would be better for this conversion to happen on the VMIRelease class so that you can be confident it's always uppercase no matter how the object was constructed?

src/pushsource/_impl/backend/koji_source.py
rohanpm: I don't think you could sensibly make this handle all types of 'image' archives.

I can't find an example build right now, but I'm pretty sure there are some builds having 'image' archives completely unrelated to container images and VM images.  Even if I'm wrong and it'd work today, it wouldn't be forwards-compatible in the case where new builds start to appear with types of 'image' not supported by this code.

rohanpm (changes requested):

docs/sources/koji.rst
```diff
@@ -77,6 +78,21 @@ Push items produced in this case may include:
 - :class:`~pushsource.OperatorManifestPushItem` - if images have attached operator
   manifest archives
 
+Acessing virtual machine images
```
rohanpm, line 81:
```suggestion
Accessing virtual machine images
```

docs/sources/koji.rst
lslebodn: I check fedora koji and was not able to find something which would look like from ImageBuilder for AWS or Azure
* https://koji.fedoraproject.org/koji/builds?start=0&order=-build_id&type=image

docs/sources/koji.rst
lslebodn: OSBS builds just container images
* https://github.com/containerbuildsystem

They have good upstream documentation about metadata
* https://github.com/containerbuildsystem/atomic-reactor/blob/master/docs/koji.md

But AMI, VHD and image for google cloud shoudl be build by image builder

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -509,6 +521,94 @@ def _push_items_from_container_build(self, nvr, meta):
...
+
+        # VMI has the {'typeinfo': {'image':{}}} on extra
+        extra = meta.get("extra") or {}
+        typeinfo = extra.get("typeinfo") or {}
+        image = typeinfo.get("image")
+
+        if image is None:
+            message = "Build %s not recognized as a virtual machine image build" % nvr
+            LOG.debug(message)
+            raise ValueError(message)
```
lslebodn, line 540: It would be nice to filter out container builds here as well.
* `meta['container_koji_task_id']` is available just for container builds
* https://github.com/containerbuildsystem/atomic-reactor/blob/master/docs/koji.md?plain=1#L79

I know ideal would be if could images has deterministic filed in extra but one cannot rely on `{'typeinfo': {'image':{}}}` yet.

It might simplify additional logic for archives.

JAVGan: Oh, thank you very much for letting me know this!

I was trying to filter out by the extension (tar) which isn't a great way to do it, so I'll adjust

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -168,6 +174,11 @@ def __init__(
                 build(s), along with associated artifacts such as operator manifest
                 bundles.
 
+            vmi_build (list[str, int])
+                Virtual Machine Image identifier(s). Can be koji IDs (integers) or
```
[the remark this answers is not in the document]
JAVGan: sure, I'll change the description, thanks!

docs/sources/koji.rst
JAVGan: Good catch!
I'll fix it

src/pushsource/_impl/backend/koji_source.py
JAVGan: Originally I thought we would be using the data from release during some steps for publishing but it seems that this is just used for RHSM which isn't something we will handle in our workflows.

Anyway, for our PoV the product should be like "RHEL" or "RHCOS" so it comes as lowercase from koji (like "rhel-azure") and then got just the first part upper case.

I liked the suggestion to do the uppercase conversion in VMIRelease so I'll adjust it there, thanks for the suggestion

tests/koji/test_koji_vmi.py
JAVGan: Whoa, I forgot!  Sure, I'll do it

tests/koji/test_koji_vmi.py
JAVGan: Ok, I'll change it

src/pushsource/_impl/backend/koji_source.py
JAVGan: Sure.

src/pushsource/_impl/backend/koji_source.py
JAVGan: Thanks!

src/pushsource/_impl/backend/koji_source.py
JAVGan: Oh, ok! Thanks

docs/sources/koji.rst
JAVGan: I'll adjust it to ImageBuilder, thanks!

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -509,6 +521,94 @@ def _push_items_from_container_build(self, nvr, meta):
...
+            cloud_types.setdefault(k, AmiPushItem)
+
+        for k in azure_types:
+            cloud_types.setdefault(k, VHDPushItem)
+
+        # Collect all image archives which are not containers (tar)
+        vmi_archives = [
+            elem
+            for elem in archives
+            if elem.get("btype") == "image" and elem.get("type_name") != "tar"
```
lslebodn, line 561: That would much even qcow2 `'type_extensions': 'qcow2'`
Do we want to support it or just: "raw", "raw-xz", "vhd", "vhd-compressed"
IMHO it is better to support just known extensions and expends it later if needed.
@rohanpm  WDYT?

src/pushsource/_impl/backend/koji_source.py
JAVGan: Just a note here while I was trying to adjust this:

By using an uppercase converter in the attribute `product` from VMIRelease it breaks the tests `test_source_against_baseline[staged-simple-ami-bc, staged-simple-ami]` with:

```
E       AssertionError: Output differs from baseline:
E       
E       --- 
E       +++ 
E       @@ -29,7 +29,7 @@
E              base_product: null
E              base_version: null
E              date: 2020-05-11
E       -      product: Fake-Product
E       +      product: FAKE-PRODUCT
E              respin: 1
E              type: ga
E              variant: Fake-Variant
```

I even tried to redefine the attribute product just for AmiRelease but that doesn't seem to work. With this I'm letting the uppercase to be in KojiSource for a while until we got a better solution

tests/koji/test_koji_vmi.py
JAVGan: Sure!

src/pushsource/_impl/backend/koji_source.py
JAVGan: @rohanpm I've changed it to consider whether `docker` is present in extra, as well as verifying whether `container_koji_task_id` is not set, as suggested by @lslebodn .

PTAL again

docs/sources/koji.rst
JAVGan: Fixed, PTAL again

src/pushsource/_impl/backend/koji_source.py
JAVGan: Fixed, PTAL again

docs/sources/koji.rst
JAVGan: I've changed it to Image Builder but IDK if there's a link to it that we can include in documentation.

Please let me know if it exists some online documentation for it, otherwise it will be just plain text.

src/pushsource/_impl/backend/koji_source.py
JAVGan: I've added a comment on it explaining what I'm doing, please let me know if it's ok

tests/koji/test_koji_vmi.py
JAVGan: Fixed! It makes more sense to hardcode as you suggested, thanks a lot!

tests/koji/test_koji_vmi.py
JAVGan: Fixed! It makes more sense to hardcode as you suggested, thanks a lot!

docs/sources/koji.rst
JAVGan: Fixed

src/pushsource/_impl/backend/koji_source.py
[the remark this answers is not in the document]
JAVGan: @lslebodn I've changed to koji build IDs but it got kinda different from other fields, I hope it's not a problem...

src/pushsource/_impl/backend/koji_source.py
JAVGan: We can do it or leave it open.. By doing it we'll need to include GCP later on and adjust the library for each new "image type" that appears, which can be a bit cumbersome IMO.

By not filtering we allow all image types and may filter them even more inside pubtools, but up to you guys, whatever works best 😃

docs/sources/koji.rst
lslebodn: TYVM for change. And just for the record (your info)
It is following github organization for the upstream
* https://github.com/osbuild

JAVGan: Do you think it is a good idea to rename to `osbuild` or at least point a link to https://www.osbuild.org/ in the docs as well?

tests/koji/test_koji_vmi.py
```diff
@@ -0,0 +1,226 @@
...
+    assert "The build foobuild-1.0-1 is for container images." in str(exc_info.value)
+
+
+def test_koji_vmis(fake_koji, koji_dir):
+    """Koji source yields requested Virtual Machine Images."""
+
+    nvr = "foobuild-1.0-1"
+    name = "foobuild"
+    version = "1.0"
+    release = "1"
```
lslebodn, line 84: a nitpick
I would keep separate name version and release and merge them to nvr
```
nvr = "%s-%s-%s" % (name, version, release)
```
later can be changed to f-string after dropping support for old version of python
I would just like to avoid copy&paste issues/typos.

JAVGan: Sure, I've fixed it now. Thanks!

docs/sources/koji.rst
lslebodn: I asked one of developers and `osbuild` is just a name of one component.
And "image builder" is a suite of projects which allows to build image as a service

And they use "image builder" in case of koji integration in the docs which you shared.
* https://www.osbuild.org/guides/image-builder-service/image-builder-koji.html

JAVGan: Thanks!

docs/sources/koji.rst
rohanpm: Is it this? https://www.osbuild.org/guides/image-builder-service/image-builder-koji.html

Also, even if there is no real example we can link to in Koji now, can we please fix the vmi_build parameter to be more accurate?

I started this thread because "fedora-azure-37-1.7.x86_64.vhd.xz" looks like the name of an archive attached to a koji build, but the code expects a build NVR or ID, not a filename.  So the example should instead look something like "fedora-azure-37-1.7", which is a feasible build NVR, shouldn't it?

JAVGan:
Note: `py27` failed due to an error on GitHub actions, I believe it will get rid by retrying the action as it's working fine in my local

docs/sources/koji.rst
```diff
@@ -77,6 +78,21 @@ Push items produced in this case may include:
...
   manifest archives
 
+Accessing virtual machine images
+...............................
+
+Use the ``vmi_build`` parameter to request virtual machine images from one or
+more builds. Builds can be specified by NVR. Builds should be produced by
+`Image Builder`_ or should use compatible metadata.
+
+``koji:https://koji.fedoraproject.org/kojihub?vmi_build=fedora-azure-37-1.7.x86_64``
```
rohanpm, line 88: I think the arch should still not be there, right?  In practice does arch usually form part of the "release"?  

It doesn't seem like it; for example, I found a build named `rhel-azure-7.9-20230110.sp.22`.

```suggestion
``koji:https://koji.fedoraproject.org/kojihub?vmi_build=fedora-azure-37-1.7``
```

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -509,6 +519,104 @@ def _push_items_from_container_build(self, nvr, meta):
...
+        azure_types = ["vhd", "vhd-compressed"]
+        cloud_types = {}
+
+        for k in ami_types:
+            cloud_types.setdefault(k, AmiPushItem)
+
+        for k in azure_types:
+            cloud_types.setdefault(k, VHDPushItem)
+
+        # Collect all image archives which are not containers
```
rohanpm, line 560: It was raised earlier if it really makes sense to assume all of these archives are VM images.  Github marked that thread as Outdated but it still seems relevant.

Can we please not do this?  I don't know a lot about koji, but I get the sense there are not very strict rules around what is recorded as an "image".  I see no reason it would be valid to assume that all images are VM images by default.  There are already counterexamples:

- the container case (which is already handled in the code)
- CD/DVD ISO images used for installers, e.g. https://url.corp.redhat.com/76d5209

Though you *can* try to write code of the form "assume they're all VM images except those we explicitly filter out", as currently done for containers: even if such code would work today, on the koji side there is apparently no requirement that 'image' archives are VM images.  So new types of image archives could be added at any time, making the code suddenly become wrong.

I think it should be limited to only those you have some more specific reason to believe are VM images.  If we can only do that by matching file extensions or similar, that's unfortunate but still better than assuming everything is a VM image.

Note, it's always easier to add support for more types later (which is backwards-compatible) than it is to remove something which shouldn't have been supported (which is potentially not backwards-compatible).  It's wise to start with supporting a narrow set of inputs and broaden it as we learn more.

rohanpm (changes requested):

docs/sources/koji.rst
JAVGan: I've added your suggestion, PTAL again

src/pushsource/_impl/backend/koji_source.py
JAVGan: Thanks for bringing this out again, IDK why it was marked as `outdated` as well.

I've updated the code to only match the known extensions from `type_name` through `cloud_types.keys()`. With this we should only include the expected VM images.

PTAL again

rohanpm (approved):

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -509,6 +519,104 @@ def _push_items_from_container_build(self, nvr, meta):
...
+                product = "-".join(splitted_name[:-1])
+            else:
+                product = splitted_name[0]
+
+            release = rel_klass(
+                arch=arch,
+                date=(meta.get("completion_time") or "").split(" ")[0],
+                product=product.upper(),
+                version=meta["version"],
+                respin=int(meta["release"]),
```
rohanpm, line 598: Are you sure that `meta["release"]` is always an integer?

Did anyone provide a link to an example build?  I'm looking at https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=2350393 and release is not an integer there, which causes this to crash:

```
$ pushsource-ls 'koji:https://brewhub.engineering.redhat.com/brewhub/?vmi_build=rhel-azure-7.9-20230110.sp.22'
[...]
  File "/home/rmcgover/src/pushsource/src/pushsource/_impl/backend/koji_source.py", line 598, in _push_items_from_vmi_build
    respin=int(meta["release"]),
           ^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: '20230110.sp.22'
```

The code should be usable with that build, right? (or can you provide at least one real build it should work against?)

rohanpm (changes requested):

rohanpm:
As well as the inline comment, python2 support was dropped in this repo, so you could consider rebasing this to remove the python2 CI failures.

src/pushsource/_impl/backend/koji_source.py
JAVGan: You're right, this is a valid `release`.

I'll work on fixing this, thanks a lot!

JAVGan: I've changed it to load the integer from just the first part of the dots if it exists, like `20230201.sp.22` would result into `20230201`.

If it's not possible to convert it to `int` it will be `0` so the program shouldn't fail again.

I also added this string into the tests to cover it, please let me know wdyt.

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -504,6 +514,108 @@ def _push_items_from_container_build(self, nvr, meta):
...
+            # marketpalce name, like "rhel-azure" or "rhel-sap-azure".
+            # We just want to get the product name without the cloud name,
+            # like "rhel" and "rhel-sap" in our examples respectively.
+            splitted_name = meta["name"].split("-")
+            if len(splitted_name) > 1:
+                product = "-".join(splitted_name[:-1])
+            else:
+                product = splitted_name[0]
+
+            # Try to get the "int" part of release for respin
```
rohanpm, line 588: Hmm, I suppose this won't crash now, but does it make sense to use this value as "respin" really?

In the example given previously, the value would be 20230201, that doesn't seem like it fits what a "respin" is documented to be.  The doc string explicitly says "Respin count. 0 for original build of an image, incremented by one for each rebuild", but the image wasn't rebuilt twenty million times.

If there isn't a reasonable value here which actually maps to "respin", I think it'd be better to just set it as 0 always, or even update the model to allow respin to be None.  I note that the full release info is still present in the KojiBuildInfo attached to the item, so leaving this out won't make anything technically impossible in your code using the items, as if you wanted to make use of the release in some way you'll still be able to do that.

rohanpm (changes requested):

src/pushsource/_impl/backend/koji_source.py
JAVGan: I agree with you.

I've changed it to pick from "release" if it's an integer or set it as 0 if not.

rohanpm (approved):

rohanpm:
It looks good to me now.

Can you please also sanity check the output against your expectations?  Here's an example of what it produces for a real build (the one mentioned earlier):

```
$ pushsource-ls 'brew:vmi_build=rhel-azure-7.9-20230110.sp.22'

# Loaded source brew:vmi_build=rhel-azure-7.9-20230110.sp.22
VHDPushItem(
    name="rhel-azure-7.9-20230110.sp.22.x86_64.vhd.xz",
    state="PENDING",
    src="/mnt/brew/packages/rhel-azure/7.9/20230110.sp.22/images/rhel-azure-7.9-20230110.sp.22.x86_64.vhd.xz",
    dest=[],
    md5sum="0546012a322d8852e18bbd492e669b16",
    sha256sum=None,
    origin=None,
    build="rhel-azure-7.9-20230110.sp.22",
    build_info=KojiBuildInfo(
        name="rhel-azure", version="7.9", release="20230110.sp.22"
    ),
    signing_key=None,
    release=VMIRelease(
        product="RHEL",
        date=datetime.date(2023, 1, 30),
        arch="x86_64",
        respin=0,
        version="7.9",
        base_product=None,
        base_version=None,
        variant=None,
        type=None,
    ),
    description="",
    generation="V2",
    sku_id=None,
    support_legacy=False,
    legacy_sku_id=None,
    disk_version=None,
    recommended_sizes=[],
    sas_uri=None,
),

# 1 item(s) found in source
```

Do the fields all look like you want them to be?  (Keeping in mind some fields will later be adjusted after the initial load e.g. using StArMap, at least that is my understanding of the plan.)

Also, is there any reason to hold back on releasing this?

JAVGan:
> Can you please also sanity check the output against your expectations? Here's an example of what it produces for a real build (the one mentioned earlier):

Sure, the output looks as expected.

> Do the fields all look like you want them to be? (Keeping in mind some fields will later be adjusted after the initial load e.g. using StArMap, at least that is my understanding of the plan.)

For me yes, @lslebodn do you have something else in mind?

> Also, is there any reason to hold back on releasing this?

I'm just waiting a final review from @lslebodn and once he gives +1 you can merge and release, nothing holding back except his +1 as well.

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -504,6 +514,108 @@ def _push_items_from_container_build(self, nvr, meta):
...
 
+    def _push_items_from_vmi_build(self, nvr, meta):
+        LOG.debug("Looking for virtual machine images on %s, %s", nvr, meta)
+
+        if not meta:
+            message = "Virtual machine image build not found in koji: %s" % nvr
+            LOG.debug(message)
+            raise ValueError(message)
+
+        if meta.get("container_koji_task_id"):
```
lslebodn, line 525: container_koji_task_id is in extra :-)
* https://github.com/containerbuildsystem/atomic-reactor/blob/master/docs/koji.md?plain=1#L79

Sorry I wrote wrong example in the other comment

lslebodn: IMHO we can even  add link to the OSBS documentation for container_koji_task_id as a comment.

JAVGan: Comment added

lslebodn (approved):
LGTM

JAVGan:
@rohanpm please help us merging and releasing whenever possible

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -504,6 +514,110 @@ def _push_items_from_container_build(self, nvr, meta):
...
+            LOG.debug(message)
+            raise ValueError(message)
+
+        # VMI has the {'typeinfo': {'image':{}}} on extra
+        extra = meta.get("extra") or {}
+        typeinfo = extra.get("typeinfo") or {}
+        image = typeinfo.get("image")
+
+        # Container images have the key `container_koji_task_id` set on extras:
+        # https://github.com/containerbuildsystem/atomic-reactor/blob/master/docs/koji.md?plain=1#L79
```
rohanpm, line 531: Minor: I think you should be linking to a specific commit or tag.  You have linked to the file in 'master' branch which can be changed at any time.  As soon as anyone inserts more lines prior to line 79, or renames or deletes the file, the link will be broken (and there will not even be an entirely reliable way to find out what this link was supposed to point to since there is no way to determine what 'master' in this repo pointed to at the time this commit was created).

rohanpm (approved):

src/pushsource/_impl/backend/koji_source.py
lslebodn: > and there will not even be an entirely reliable way to find out what this link was supposed to point to

It is true that commit are better but there is a reliable way how to get get to the correct line.
One would need to check date when the comment was added here(elease-engineering/pushsource) and use git "archeology" to checkout state of referenced repository at that date. It is not super ideal UX and also rely on disabled force push to master/main branch (which is de-facto standard in most of upstream projects)

On the other hand replacing  branch with commit is trivial as well `s!blob/master!blob/87155e0829b244192201b16ddbb693da6cea7811!`

rohanpm: > there is a reliable way how to get get to the correct line. One would need to check date when the comment was added here(elease-engineering/pushsource) and use git "archeology" to checkout state of referenced repository at that date. 

How would you do it?  I can obtain the commit or author date of this commit in pushsource repo.  Wouldn't I then need github to tell me, "in containerbuildsystem/atomic-reactor, at time X, branch `master` pointed at commit Y" ?

I am aware that:

- commits have a commit date and author date, but this has no direct relationship to when a remote branch was pointing at a particular commit.
- locally, git reflog knows the history of the refs in my repo.  It can tell me what `master` pointed to at time X.  But this is only my local refs, which again has no direct relationship to the remote branches.

If you do know a reliable way to figure out what a remote repo's branch was pointing to at a specific time, please share as I'm interested to learn it.

PR #295, opened 2023-01-05
Make pushsource compatible with attrs version 22.2.0

querti:
Since the attrs version release, pub tests have begun failing with the error "TypeError: keywords must be strings". Attrs have added an "alias" parameter to each attribute[1], which is used in the attrs's "evolve" method to construct a new instance. This causes issues with the attribute "from", which has some special logic implemented in order to be usable at all[2]. The attribute "from" is generated from "from_", copying all its parameters. Since "alias" is a new parameter, it isn't copied and thus its value is set no "None". attrs's "evolve" method uses **kwargs to create a new instance, and it sets one of the keys in **kwargs to "None" (from alias). Keys in **kwargs must be strings, which causes the TypeError. It can be fixed by also setting "alias" in the newly created "from" attribute. In case of "from_", it has the same value as "name" parameter ("from_"), which is why we don't want to copy it from the old attribute, but explicitly set it to "from" (otherwise it would be "from_"). The change should be backwards compatible with older versions of attrs.

[1] https://github.com/python-attrs/attrs/pull/950
[2] https://github.com/release-engineering/pushsource/pull/108

Refers to CLOUDDST-17044


querti:
@release-engineering/distribution pls review

rohanpm (approved):

midnightercz (approved):

PR #296, opened 2023-01-06
Poll for staged content

querti:
Staged content is checked before the main polling mechanism, meaning that PushStaged code fails unless the staged content is present in the snapmirror. Add the polling mechanism to staged content. If staged path doesn't exist in the mounted volume, its existence will be polled until it appears. In this scenario it is safe to assume that the staged content is new and up-to-date.

In rare cases, existing staged content may be altered and pushed again. Since pushsource cannot know if the altered content has been propagated to snapmirror, it has to wait the full poll_timeout period to be safe. Thus, if the content already exists on the volume, wait the full period to ensure it's up to date.

This implementation assumes that push-staged is ran with only one path. Otherwise, the code will wait the full poll_timeout for each path, which is unnecessary and time inefficient. Running push-staged with multiple paths is apparently not a real use-case, so it should be fine in 99% of cases.

Refers to CLOUDDST-17045

### DIFF
--- a/src/pushsource/_impl/backend/staged/staged_source.py
+++ b/src/pushsource/_impl/backend/staged/staged_source.py
@@ -143,7 +143,9 @@ class StagedSource(
         # Apply wait only if topdir mtime is less than X seconds older than current time.
         # The idea is "Only apply wait if the staging dir was recently modified".
         # The lower the value, the fewer pushes will wait,the higher the probability of missing data
-        apply_wait_topdir_age = int(os.getenv("PUSHSOURCE_APPLY_WAIT_TOPDIR_AGE") or "0")
+        apply_wait_topdir_age = int(
+            os.getenv("PUSHSOURCE_APPLY_WAIT_TOPDIR_AGE") or "0"
+        )
 
         if not os.path.exists(topdir):
             LOG.info(
--- a/src/pushsource/_impl/backend/staged/staged_source.py
+++ b/src/pushsource/_impl/backend/staged/staged_source.py
@@ -3,6 +3,8 @@ import threading
 import logging
 import itertools
 import functools
+import time
+from datetime import datetime
 
 import yaml
 import json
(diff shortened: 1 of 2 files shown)


src/pushsource/_impl/backend/staged/staged_source.py
```diff
@@ -136,6 +137,25 @@ def _push_items_for_leafdir(self, leafdir, metadata):
     def _push_items_for_topdir(self, topdir):
         LOG.info("Checking files in: %s", topdir)
 
+        timeout = int(os.getenv("PUSHSOURCE_SRC_POLL_TIMEOUT") or "0")
+        poll_rate = int(os.getenv("PUSHSOURCE_SRC_POLL_RATE") or "30")
+
+        # If staged path doesn't exists, we can assume that once it does, it will
+        # contain fresh content. It's sufficient to only wait until it appears.
+        if not os.path.exists(topdir):
+            wait_exist(topdir, timeout, poll_rate)
```
rohanpm, line 146: I don't know exactly how the replication mechanism works, but this logic seems a bit suspicious because it will only work under the assumption that, if a parent directory appears, all subdirectories have been fully replicated.  It does not seem a safe assumption without diving into exactly how it works.

I mean for example, looking at https://release-engineering.github.io/pushsource/sources/staged.html#staging-structure , can it be the case that:

- initially "root/" does not exist
- we poll for it to exist, and find that it eventually appears
- then we proceed through the `for entry in scandir(topdir)` link below
- but only half of the 'destination' dirs have replicated, so scandir misses half of the staged content

Is it possible to get a meaningful `mtime` of the topdir?  If so, the logic could be something like:

Check the mtime.  Sleep until the current time is at least (mtime + timeout).  Imagining a timeout of 15 minutes, it means if a directory was initially replicated just as we get up to this code, then we'll wait 15 minutes to hopefully ensure that all of the subdirectories are also fully replicated.

src/pushsource/_impl/backend/staged/staged_source.py
```diff
@@ -136,6 +137,25 @@ def _push_items_for_leafdir(self, leafdir, metadata):
...
+        # If the path already exists, it may have been modified which means that pushsource
```
rohanpm, line 147: I'm curious whether this has been inspired by any real problems encountered yet, or is it only anticipating problems?

I'm not sure that we should add this.  I think the users should be strongly discouraged from modifying staging dirs after creation, and they should instead create new ones (or at least a new *path*, e.g. by renaming the dir).

Mutable staging dirs would always have caused some problems even before recent Pub changes:

- the filesystem was always synced to multiple regions with some delay. So by modifying the staging area after creation, they've always been creating multiple inconsistent views of the staging area.
- if staging area is modified after initial use, it becomes impossible to reproduce what Pub did in a given push, and difficult to debug if something unexpected happened.

Basically, if path X is allowed to contain different content at different times, and there is caching or replication in play (and there always has been), then life is harder for everyone.  So the users shouldn't be doing that.

It seems like this change is going to impose a severe penalty on almost *all* users of push-staged in order to make life easier for those users doing something they shouldn't be doing it in the first place.

It doesn't seem right to me.  How about just telling the users/letting them learn that, if they need to change a staging area after creation, it's better to create a new one, rename the dir, or wait 15 minutes?  This means the users who are modifying their staging areas will feel the pain and be encouraged to change their processes, rather than *all* the users feeling pain even if they're using the system in an ideal way.

rohanpm (changes requested):

src/pushsource/_impl/backend/staged/staged_source.py
midnightercz: I think in the past it often happen there was typo in repo directory or missing product ids. I'm not sure how often it was fixed in place or how often new directory was created. But I assume relengs usually choose the easiest way.

src/pushsource/_impl/backend/staged/staged_source.py
rohanpm: I don't like the idea that every push-staged would have a 15 minutes delay at the beginning even if, for example, the user is pushing a staging area to production which was already pushed to stage hours/days earlier (which I think is a common case).  We'd probably get bug reports/complaints about that straight away.

How about this logic then:

- check the mtime of the topdir
- if and only if mtime falls in the range of (current_time - X minutes) .. (current_time), then apply the delay
- the value of X is configurable by env var

If we start by setting X to 24 hours, then the overall behavior will be: if we process a dir created within the last day, we need to delay because we assume it might still be changing.  If we process a dir created earlier than that, we assume it's stable.  And since it's configurable, if we chose a bad value of X we can tune it without code changes.  Putting a very large number will get us back to "delay every time" if it turns out this is really necessary.

Also, if we're doing anything along those lines it'd probably be best to have an INFO-level log message with the mtime of the directory, so that we can later reason about how much it should have delayed.

querti: Changed the logic to your proposal.

rohanpm (approved):

emilyzheng (approved):

PR #297, opened 2023-01-18
Introduce pub source backend [RHELDST-7514]

rbikar:
Introduced new pushsource backend - `pub` for extracting push items
from Pub push item list by task id.

Pub backend currently supports extraction of AMI push items only. They
are extracted from images.json file stored on Pub service for each (AMI) task.

Little addition to AmiPushItem model class - added `image_id` attribute
that is unique identifier of AMI in AWS that is only knonw, if the AMI was
pushed previously to AWS.

### DIFF
--- a/docs/sources/pub.rst
+++ b/docs/sources/pub.rst
@@ -1,0 +1,30 @@
+Source: pub
+==============
+
+The ``pub`` push source allows the loading of content from an instance of Pub.
+
+Supported content types:
+
+* AMIs
+
+
+pub source URLs
+------------------
+
+The base form of an pub source URL is:
+
+``pub:pub-url?task_id=100[,200[,...]]``
+
+For example, referencing a single task would look like:
+
+``pub:https://pub.example.com?task_id=1234``
+
+The provided Pub URL should be the base URL of the service,
+and not any specific API endpoint.
+
+Python API reference
+--------------------
+
+.. autoclass:: pushsource.PubSource
+   :members:
+   :special-members: __init__
--- a/src/pushsource/_impl/backend/pub_source/pub_client.py
+++ b/src/pushsource/_impl/backend/pub_source/pub_client.py
@@ -1,0 +1,63 @@
+import json
+import logging
+import os
+import threading
+
+
+import requests
+from more_executors import Executors
+
+
+LOG = logging.getLogger("pushsource.pub_client")
+
+
+class PubClient(object):
+    def __init__(self, threads, url, **retry_args):
+        self._executor = (
+            Executors.thread_pool(name="pushsource-pub-client", max_workers=threads)
+            .with_map(self._unpack_response)
+            .with_retry(**retry_args)
+            .with_cancel_on_shutdown()
+        )
+        self._url = url
+        self._tls = threading.local()
+
+    def shutdown(self):
+        self._executor.shutdown(True)
+
+    def _unpack_response(self, response):
+        try:
+            response.raise_for_status()
+            out = response.json()
+        except json.JSONDecodeError:
+            # do not treat invalid json as fatal error
+            out = None
+        return out
+
+    @property
+    def _session(self):
+        if not hasattr(self._tls, "session"):
+            LOG.debug(
+                "Creating requests Session for client of Pub service: %s", self._url
+            )
+            self._tls.session = requests.Session()
+        return self._tls.session
+
+    def _do_request(self, **kwargs):
+        return self._session.request(**kwargs)
+
+    def get_ami_json_f(self, task_id):
+        """
+        Returns Future[dict|list] holding json obj with AMI push items returned from Pub for given task id.
+        """
+        endpoint = "pub/task"
+        url_ending = "log/images.json"
+        params = {"format": "raw"}
+        url = os.path.join(self._url, endpoint, str(task_id), url_ending)
+
+        LOG.info("Requesting Pub service for %s of task: %s", url_ending, str(task_id))
+        ft = self._executor.submit(
+            self._do_request, method="GET", url=url, params=params
+        )
+
+        return ft
--- a/src/pushsource/_impl/backend/pub_source/pub_source.py
+++ b/src/pushsource/_impl/backend/pub_source/pub_source.py
@@ -1,0 +1,106 @@
+import logging
+from concurrent import futures
+
+from more_executors import Executors
+
+from ...helpers import (
+    as_completed_with_timeout_reset,
+    force_https,
+    list_argument,
+    try_int,
+)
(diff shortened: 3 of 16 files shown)


docs/sources/pub.rst
```diff
@@ -0,0 +1,30 @@
...
+pub source URLs
+------------------
+
+The base form of an pub source URL is:
+
+``pub:pub-url?task_id=100[,200[,...]]``
+
+For example, referencing a single task would look like:
+
+``errata:https://pub.example.com?task_id=1234``
```
rohanpm, line 20:
```suggestion
``pub:https://pub.example.com?task_id=1234``
```

src/pushsource/_impl/backend/pub_source/pub_source.py
```diff
@@ -0,0 +1,108 @@
...
+                "https://pub.example.com:8123".
+
+            task_id (int, str, list[str], list[int])
+                Task ID(s) to be used as push item source.
+                If a single string is given, multiple IDs may be
+                comma-separated.
+
+            threads (int)
+                Number of threads used for concurrent queries to Errata Tool
+                and koji.
```
rohanpm, line 43: ET and koji are not involved, so it should say either "to Pub" or maybe just not specify any system.

```suggestion
                Number of threads used for concurrent queries.
```

src/pushsource/_impl/backend/pub_source/pub_source.py
```diff
@@ -0,0 +1,108 @@
...
+
+            timeout (int)
+                Number of seconds after which an error is raised, if no progress is
+                made during queries to Errata Tool.
```
rohanpm, line 47:
```suggestion
                made during queries.
```

(or "...to Pub")

src/pushsource/_impl/model/ami.py
```diff
@@ -13,6 +13,23 @@ class AmiRelease(VMIRelease):
     for backwards compatibility.
     """
 
+    @classmethod
+    def from_data(cls, data):
```
rohanpm, line 17: Like it is with erratum.py, can we please name these `_from_data` so they're considered private?  We don't want other codebases to start relying on the dict structure.

src/pushsource/_impl/model/ami.py
```diff
@@ -71,3 +98,36 @@ class AmiPushItem(VMIPushItem):
         validator=optional(instance_of(AmiBillingCodes)),
     )
     """Billing codes associated with this image."""
+
+    image_id = attr.ib(type=str, default=None, validator=optional_str)
+    """AMI Image ID that is reference for the image in AWS."""
```
rohanpm, line 103:
```suggestion
    """AMI Image ID used to reference the image in AWS."""
```

tests/pub/test_pub_amis.py
```diff
@@ -0,0 +1,300 @@
...
+    with Source.get("pub:%s" % pub_url, task_id=task_id) as source:
+        push_items = [item for item in source]
+
+    # no exception raise but also no push items
+    assert len(push_items) == 0
+    # with follow line captured in log
+    assert caplog.messages[0] == "Invalid task_id: a7df9"
+
+
+def test_pub_source_empty_response(requests_mock, caplog):
```
rohanpm, line 258: I don't think it should behave this way actually.  Can we please make it so that the absence of images.json is not considered an error?  If there are no AMIs I think it should just successfully produce nothing.

In general, source behavior is like this: get and return all items of all supported types found on the referenced object (staging directory; ET advisory; koji build).  It's not an error if any specific type has no content, or even if all types have no content.  I know "all supported types" for pub backend right now is only AMIs, but it still makes sense for the behavior to be consistent with other backends and to be consistent with future behavior of pub backend once other types are also supported.

If you're worried about users giving a wrong task ID and not noticing that there are no items produced, it could log a warning in the case where no items are found.

rohanpm (changes requested):

rbikar:
Fixed review issues + py27 CI will be dropped by https://github.com/release-engineering/pushsource/pull/300

src/pushsource/_impl/backend/pub_source/pub_client.py
```diff
@@ -0,0 +1,66 @@
+import json
+import logging
+import os
+import threading
+import six
+
+import requests
+from more_executors import Executors
+
+if six.PY2:  # pragma: no cover
```
rohanpm, line 10: I guess this will conflict now as the "remove py2 support" change was submitted, which removes six.

src/pushsource/_impl/backend/pub_source/pub_source.py
```diff
@@ -0,0 +1,106 @@
...
+            name="pushsource-pub", max_workers=threads
+        ).with_cancel_on_shutdown()
+
+    def _validate_task_ids(self, task_ids):
+        out = []
+        for task_id in task_ids:
+            if isinstance(task_id, int):
+                out.append(task_id)
+            else:
+                LOG.warning("Invalid task_id: %s", task_id)
```
rohanpm, line 66: This should probably mention Pub somewhere.  In a complex task with a lot of log messages it seems like it might not be obvious this is referring to a Pub task ID.

src/pushsource/_impl/backend/pub_source/pub_source.py
```diff
@@ -0,0 +1,106 @@
...
+            push_items = AmiPushItem._from_data(json_obj)
+            out.extend(list_argument(push_items))
+        except KeyError:
+            LOG.warning("Unsupported or unparsable push item: %s", str(json_obj))
+
+        return out
+
+    def __iter__(self):
+        # Get json Pub responses for all task ids
+        json_fs = [self._client.get_json_f(task_id) for task_id in self._task_ids]
```
rohanpm, line 89: It looks like all the code around here is going to work fine, but personally I find the naming misleading.

The function names like get_json_f and _push_items_from_json give the impression it's generic and could later be extended for other types, but in reality it's AMI-specific and couldn't be extended to other types cleanly.  For example, if a later content type needs us to start fetching pub-mapfile.json, is get_json_f going to start returning two JSON documents combined and then _push_items_from_json understands how to unpack those?  Would the methods have to be renamed or refactored at that point?

I feel like the way things are split up/named in koji_source is a better approach, which would be something more like `get_ami_json_f` (or maybe `get_json_f("log/images.json")`) followed by `_push_items_from_ami_json`.  Then the natural way to extend it for later types is to add new similar methods which each focus on their supported type.

rbikar:
* rebased on current master
* fixed naming of methods and logging

rohanpm (approved):

PR #301, opened 2023-01-30
Fix bad validator on VHDPushItem.legacy_sku_id

JAVGan:
This PR fixes a bad behaviro on `VHDPushItem.legacy_sku_id' which should only be set whenever `support_legacy` is True.

If the constraint is respected the value of `legacy_sku_id` can be any string, however the previous version was limiting it in `V1` or `V2` (same as `generation`) which shouldn't be the case.


JAVGan:
@rohanpm @lslebodn PTAL, I caught this while implementing our new _pubtool_.

I believe it's good to have this fix included together with #294 in the same release.

rohanpm (approved):

PR #304, opened 2023-02-06
additions to Pushsource's AMI metadata

ashwini3326:
While working through adding publishing a product version to AWS it was discovered we need more metadata to complete it. Following are the new metadata which would be required for the same. @JAVGan @jajreidy @lslebodn PTAL

### DIFF
--- a/src/pushsource/_impl/model/ami.py
+++ b/src/pushsource/_impl/model/ami.py
@@ -56,6 +56,43 @@ class AmiBillingCodes(object):
 
 
 @attr.s()
+class AmiSecurityGroup(object):
+    """Security group information to be associated with a Marketplace VM."""
+
+    from_port = attr.ib(type=int, validator=instance_of(int))
+    """If the protocol is TCP or UDP, this is the start of the port range.
+    If the protocol is ICMP or ICMPv6, this is the type number.
+    A value of -1 indicates all ICMP/ICMPv6 types. If you specify all ICMP/ICMPv6 types,
+    you must specify all ICMP/ICMPv6 codes."""
+
+    ip_protocol = attr.ib(type=str, validator=instance_of_str)
+    """The IP protocol name ( tcp , udp , icmp , icmpv6 )."""
+
+    ip_ranges = attr.ib(type=list, converter=frozenlist)
+    """List of the IPv4 ranges, for example ['22.22.22.22', '33.33.33.33'].
+
+    :type: list[str]
+    """
+
+    to_port = attr.ib(type=int, validator=instance_of(int))
+    """If the protocol is ICMP or ICMPv6, this is the code.
+    A value of -1 indicates all ICMP/ICMPv6 codes.
+    If you specify all ICMP/ICMPv6 types, you must specify all ICMP/ICMPv6 codes.."""
+
+    @classmethod
+    def _from_data(cls, data):
+        """Instantiate SecurityGroup from raw dict"""
+
+        kwargs = {
+            "from_port": data["from_port"],
+            "ip_protocol": data["ip_protocol"],
+            "ip_ranges": data["ip_ranges"],
+            "to_port": data["to_port"],
+        }
+        return cls(**kwargs)
+
+
+@attr.s()
 class AmiPushItem(VMIPushItem):
     """A :class:`~pushsource.PushItem` representing an Amazon Machine Image (or "AMI").
 
@@ -99,6 +136,21 @@ class AmiPushItem(VMIPushItem):
     )
     """Billing codes associated with this image."""
 
+    release_notes = attr.ib(type=str, default=None, validator=optional_str)
+    """"Notes regarding changes in the newest version."""
+
+    usage_instructions = attr.ib(type=str, default=None, validator=optional_str)
+    """Step by step instructions for the end-user to launch, configure and access the product.
+
+    Write with a less-technical customer in mind.
+    """
+
+    recommended_instance_type = attr.ib(type=str, default=None, validator=optional_str)
+    """Recommended instance type for AMI e.g. "t2.micro", "r6a.large" """
+
+    marketplace_entity_type = attr.ib(type=str, default=None, validator=optional_str)
+    """Type of entity e.g. "AMIProduct", "SaasProduct", "ServerProduct". """
+
     image_id = attr.ib(type=str, default=None, validator=optional_str)
     """AMI Image ID used to reference the image in AWS."""
 
@@ -106,6 +158,18 @@ class AmiPushItem(VMIPushItem):
         type=bool, default=None, validator=optional(instance_of(bool))
     )
     """``True`` if the image is allowed to be released publicly (shared with group "all")."""
+
+    scanning_port = attr.ib(type=int, default=None, validator=optional(instance_of(int)))
+    """AMI scanning port, used when importing the AMI into AWS Marketplace to validate the AMI."""
+
+    user_name = attr.ib(type=str, default=None, validator=optional_str)
+    """The username used to login to the AMI. (Usually set to ec2-user)"""
+
+    version_title = attr.ib(type=str, default=None, validator=optional_str)
+    """The title given to a version. This will display in AWS Marketplace as the name of the version."""
+
+    security_groups = attr.ib(type=list, default=attr.Factory(frozenlist), converter=frozenlist)
+    """Automatically created security groups for the product. """
 
     @classmethod
     def _from_data(cls, data):
@@ -136,6 +200,14 @@ class AmiPushItem(VMIPushItem):
             ),
             "image_id": data.get("ami") or None,
             "public_image": data.get("public_image"),
+            "release_notes": data.get("release_notes") or None,
+            "usage_instructions": data.get("usage_instructions") or None,
+            "recommended_instance_type": data.get("recommended_instance_type") or None,
+            "marketplace_entity_type": data.get("marketplace_entity_type") or None,
+            "scanning_port": data.get("scanning_port") or None,
+            "user_name": data.get("user_name") or None,
+            "version_title": data.get("version_title") or None,
+            "security_groups":[AmiSecurityGroup._from_data(security_group) for security_group in (data.get("security_groups") or []) ],
         }
 
         return cls(**kwargs)
--- a/src/pushsource/_impl/model/ami.py
+++ b/src/pushsource/_impl/model/ami.py
@@ -159,7 +159,9 @@ class AmiPushItem(VMIPushItem):
     )
     """``True`` if the image is allowed to be released publicly (shared with group "all")."""
 
-    scanning_port = attr.ib(type=int, default=None, validator=optional(instance_of(int)))
+    scanning_port = attr.ib(
+        type=int, default=None, validator=optional(instance_of(int))
+    )
     """AMI scanning port, used when importing the AMI into AWS Marketplace to validate the AMI."""
 
     user_name = attr.ib(type=str, default=None, validator=optional_str)
@@ -168,7 +170,9 @@ class AmiPushItem(VMIPushItem):
     version_title = attr.ib(type=str, default=None, validator=optional_str)
     """The title given to a version. This will display in AWS Marketplace as the name of the version."""
 
-    security_groups = attr.ib(type=list, default=attr.Factory(frozenlist), converter=frozenlist)
+    security_groups = attr.ib(
+        type=list, default=attr.Factory(frozenlist), converter=frozenlist
+    )
     """Automatically created security groups for the product. """
 
     @classmethod
@@ -207,7 +211,10 @@ class AmiPushItem(VMIPushItem):
             "scanning_port": data.get("scanning_port") or None,
             "user_name": data.get("user_name") or None,
             "version_title": data.get("version_title") or None,
-            "security_groups":[AmiSecurityGroup._from_data(security_group) for security_group in (data.get("security_groups") or []) ],
+            "security_groups": [
+                AmiSecurityGroup._from_data(security_group)
+                for security_group in (data.get("security_groups") or [])
+            ],
         }
 
         return cls(**kwargs)
--- a/src/pushsource/_impl/schema/staged-schema.yaml
+++ b/src/pushsource/_impl/schema/staged-schema.yaml
@@ -104,6 +104,66 @@ definitions:
                 type:
                 - boolean
                 - "null"
+            release_notes:
+                $ref: "#/definitions/optional_string"
+
+            usage_instructions:
+                $ref: "#/definitions/optional_string"
+
+            recommended_instance_type:
+                $ref: "#/definitions/optional_string"
+
+            marketplace_entity_type:
+                $ref: "#/definitions/optional_string"
+
+            scanning_port:
+                type:
+                    - number
+                    - "null"
+                minimum: 0
+                maximum: 65536
+
+            user_name:
+                $ref: "#/definitions/optional_string"
+            version_title:
+                $ref: "#/definitions/optional_string"
+            security_groups:
+                type:
+                    - array
+                    - "null"
+                items:
+                    type: object
+                    properties:
+                        ip_protocol:
+                            type: string
+                        ip_ranges:
+                            type: array
+                            items:
+                                type: string
+                    if:
+                        properties:
+                            ip_protocol:
+                                enum: ["icmp", "icmpv6"]
+                    then:
+                        properties:
+                            from_port:
+                                type: number
+                                minimum: -1
+                                maximum: 255
+                            to_port:
+                                type: number
+                                minimum: -1
+                                maximum: 255
+                    else:
+                        properties:
+                            from_port:
+                                type: number
+                                minimum: 0
+                                maximum: 65536
+                            to_port:
+                                type: number
+                                minimum: 0
+                                maximum: 65536
 
         required:
         - release
@@ -244,6 +304,7 @@ definitions:
         required:
         - name
         - codes
+
 
 ###############################################################################
 #  Main schema
--- a/tests/baseline/cases/staged-simple-ami-bc.yml
+++ b/tests/baseline/cases/staged-simple-ami-bc.yml
@@ -21,10 +21,12 @@ items:
     - dest1
     ena_support: true
     image_id: null
+    marketplace_entity_type: AMIProduct
     md5sum: null
     name: fake-image.raw
     origin: {{ src_dir }}/tests/staged/data/simple_ami_with_bc
     public_image: true
+    recommended_instance_type: t2.micro
     region: cn-north-1
     release:
       arch: x86_64
@@ -36,12 +38,24 @@ items:
       type: ga
       variant: Fake-Variant
       version: Fake-Version
+    release_notes: fake release notes
     root_device: /dev/sda1
+    scanning_port: 1244
+    security_groups:
+    - from_port: 0
+      ip_protocol: Fake-Protocol
+      ip_ranges:
+      - 22.22.22.22
+      - 33.33.33.33
+      to_port: 1
     sha256sum: null
     signing_key: null
     src: {{ src_dir }}/tests/staged/data/simple_ami_with_bc/dest1/AWS_IMAGES/fake-image.raw
     sriov_net_support: simple
     state: PENDING
     type: access
+    usage_instructions: fake usage instructions
+    user_name: Fake-Username
+    version_title: Fake-Version-Title
     virtualization: hvm
     volume: gp2
--- a/docs/model/ami.rst
+++ b/docs/model/ami.rst
@@ -9,3 +9,6 @@ Push items: AMI
 
 .. autoclass:: pushsource.AmiBillingCodes()
    :members:
+
+.. autoclass:: pushsource.AmiSecurityGroup()
+   :members:
--- a/src/pushsource/__init__.py
+++ b/src/pushsource/__init__.py
@@ -22,6 +22,7 @@ from pushsource._impl.model import (
     AmiPushItem,
     AmiRelease,
     AmiBillingCodes,
+    AmiSecurityGroup,
     VHDPushItem,
     ErratumPushItem,
     ErratumReference,
--- a/src/pushsource/_impl/backend/staged/staged_ami.py
+++ b/src/pushsource/_impl/backend/staged/staged_ami.py
@@ -57,6 +57,15 @@ class StagedAmiMixin(StagedBaseMixin):
             "sriov_net_support",
             "ena_support",
             "public_image",
+            "release_notes",
+            "usage_instructions",
+            "recommended_instance_type",
+            "marketplace_entity_type",
+            "image_id",
+            "scanning_port",
+            "user_name",
+            "version_title",
+            "security_groups",
         ]
 
         for key in image_attrs:
--- a/src/pushsource/_impl/model/__init__.py
+++ b/src/pushsource/_impl/model/__init__.py
@@ -22,6 +22,6 @@ from .container import (
 from .modulemd import ModuleMdPushItem, ModuleMdSourcePushItem
 from .comps import CompsXmlPushItem
 from .productid import ProductIdPushItem
-from .ami import AmiPushItem, AmiRelease, AmiBillingCodes
+from .ami import AmiPushItem, AmiRelease, AmiBillingCodes, AmiSecurityGroup
 from .azure import VHDPushItem
 from .vms import VMIPushItem, VMIRelease
--- a/tests/baseline/cases/staged-simple-ami.yml
+++ b/tests/baseline/cases/staged-simple-ami.yml
@@ -17,10 +17,12 @@ items:
     - dest1
     ena_support: true
     image_id: null
+    marketplace_entity_type: null
     md5sum: null
     name: fake-image.raw
     origin: {{ src_dir }}/tests/staged/data/simple_ami
     public_image: true
+    recommended_instance_type: null
     region: cn-north-1
     release:
       arch: x86_64
@@ -32,12 +34,18 @@ items:
       type: ga
       variant: Fake-Variant
       version: Fake-Version
+    release_notes: null
     root_device: /dev/sda1
+    scanning_port: null
+    security_groups: []
     sha256sum: null
     signing_key: null
     src: {{ src_dir }}/tests/staged/data/simple_ami/dest1/AWS_IMAGES/fake-image.raw
     sriov_net_support: simple
     state: PENDING
     type: access
+    usage_instructions: null
+    user_name: null
+    version_title: null
     virtualization: hvm
     volume: gp2
--- a/tests/pub/data/123456/images.json
+++ b/tests/pub/data/123456/images.json
@@ -32,6 +32,18 @@
       "state": "PUSHED", 
       "type": "hourly", 
       "virtualization": "hvm", 
-      "volume": "gp2"
+      "volume": "gp2",
+      "release_notes": "Fake release notes",
+      "usage_instructions": "Fake use instructions",
+      "recommended_instance_type": "t2.micro",
+      "marketplace_entity_type": "AMI",
+      "scanning_port":  10,
+      "user_name": "Fake-Username",
+      "version_title": "version title",
+      "security_groups": [
+        {"from_port":10, "ip_protocol": "udp", "ip_ranges":["8.1.8.8"], "to_port":90},
+        {"from_port":11, "ip_protocol": "tcp", "ip_ranges":["8.2.8.8"], "to_port":91}
+      ]
+
     }
 ]
--- a/tests/pub/data/200/images.json
+++ b/tests/pub/data/200/images.json
@@ -32,7 +32,9 @@
     "state": "PUSHED", 
     "type": "hourly", 
     "virtualization": "hvm", 
-    "volume": "gp2"
+    "volume": "gp2",
+    "security_groups": [{"from_port":-1, "ip_protocol": "icmpv6", "ip_ranges":["255.255.255.255"], "to_port":255}]
+
   }, 
   {
     "ami": "ami-fake-200-B", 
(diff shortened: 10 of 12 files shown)


src/pushsource/_impl/model/ami.py
```diff
@@ -71,3 +71,39 @@ class AmiPushItem(VMIPushItem):
         validator=optional(instance_of(AmiBillingCodes)),
     )
     """Billing codes associated with this image."""
+
+    image_name = attr.ib(type=str, default=None, validator=optional_str)
+    """image name  """
```
JAVGan, line 76: @jajreidy @lslebodn I believe we can get the `image_name`  from existing attributes just using something similar to [this](https://github.com/release-engineering/pubtools-ami/blob/main/pubtools/_ami/tasks/push.py#L140-L173) in our _pubtool_, no?

src/pushsource/_impl/model/ami.py
```diff
@@ -71,3 +71,39 @@ class AmiPushItem(VMIPushItem):
...
+
+    description = attr.ib(type=str, default=None, validator=optional_str)
+    """A description for the image"""
+
+    arch = attr.ib(type=str, default=None, validator=optional_str)
+    """Architecture type of the image"""
```
JAVGan, line 82: This one we can get from `AmiRelease(VMIRelease)` as defined [here](https://github.com/release-engineering/pushsource/blob/master/src/pushsource/_impl/model/vms.py#L19-L20) thus it's not necessary here IMO

src/pushsource/_impl/model/ami.py
```diff
@@ -71,3 +71,39 @@ class AmiPushItem(VMIPushItem):
...
+
+    virt_type = attr.ib(type=str, default=None, validator=optional_str)
+    """Virtualization type """
```
JAVGan, line 85: What is the difference between this and the already existing attribute `virtualization`?

What are the expected values?

src/pushsource/_impl/model/ami.py
```diff
@@ -71,3 +71,39 @@ class AmiPushItem(VMIPushItem):
...
+
+    accounts = attr.ib(type=str, default=None, validator=optional_str)
+    """Accounts """
+
+    groups = attr.ib(type=str, default=None, validator=optional_str)
+    """Groups"""
```
JAVGan, line 91: @jajreidy @lslebodn Will it be required to use multiple accounts/groups or just the internal one?

Because looking here in [pubtools-ami](https://github.com/release-engineering/pubtools-ami/blob/main/pubtools/_ami/tasks/push.py#L422-L430) they get it from command line args so maybe we can get this from the credentials file as well, instead of relying on StArMap and adding an attribute that might not be required?

Then we could have the credentials file (from args) more or less like this:

```
{
    "marketplace_account": "aws-na",
    "auth": {
        "access_id": "**************",
        "secret_key": "**************",
        "region": "**************",
        "groups": ["group1", "group2", ...],
    },
}
```

WDYT?

src/pushsource/_impl/model/ami.py
```diff
@@ -71,3 +71,39 @@ class AmiPushItem(VMIPushItem):
...
+    """Version of operating system"""
+
+    usage_instructions = attr.ib(type=str, default=None, validator=optional_str)
+    """Instructions for use of AMI"""
+
+    recommended_instance_type = attr.ib(type=str, default=None, validator=optional_str)
+    """Recommended instance type for AMI"""
+
+    product_type = attr.ib(type=str, default=None, validator=optional_str)
+    """Type of product"""
```
JAVGan, line 109: What is expected to be here? If we have some examples it would be better.

src/pushsource/_impl/model/ami.py
```diff
@@ -71,3 +71,39 @@ class AmiPushItem(VMIPushItem):
...
+    """Groups"""
+
+    release_notes = attr.ib(type=str, default=None, validator=optional_str)
+    """"Release notes for AMI """
+
+    operating_system_name = attr.ib(type=str, default=None, validator=optional_str)
+    """Name of operating system"""
+
+    operating_system_version = attr.ib(type=str, default=None, validator=optional_str)
+    """Version of operating system"""
```
JAVGan, line 100: Can't we use the [AmiRelease.version](https://github.com/release-engineering/pushsource/blob/master/src/pushsource/_impl/model/vms.py#L26-L28) for this?

src/pushsource/_impl/model/ami.py
```diff
@@ -71,3 +71,39 @@ class AmiPushItem(VMIPushItem):
...
+    """Accounts """
+
+    groups = attr.ib(type=str, default=None, validator=optional_str)
+    """Groups"""
+
+    release_notes = attr.ib(type=str, default=None, validator=optional_str)
+    """"Release notes for AMI """
+
+    operating_system_name = attr.ib(type=str, default=None, validator=optional_str)
+    """Name of operating system"""
```
JAVGan, line 97: What is expected to be here, something like "RHEL" or just "Linux"?

Depending on it we can simplify by keeping the information just in [AmiRelease.product](https://github.com/release-engineering/pushsource/blob/master/src/pushsource/_impl/model/vms.py#L12-L14)?

JAVGan (commented):
I believe some attributes are redundant or not necessary to be defined in pushsource level, but that's something we can discuss.

@jajreidy @lslebodn WDYT?

src/pushsource/_impl/model/ami.py
jajreidy: That'd probably be fine.

src/pushsource/_impl/model/ami.py
jajreidy: Yeah that makes sense.

src/pushsource/_impl/model/ami.py
jajreidy: product_type would be AMIProduct, SaasProduct, ServerProduct (which is also AMIProduct, I couldn't differentiate the two.)

src/pushsource/_impl/model/ami.py
jajreidy: virtualization might be the same input. As far as AWS is concerned these are the virtualization types:

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/virtualization_types.html

src/pushsource/_impl/model/ami.py
jajreidy: Yeah that'd make sense.

src/pushsource/_impl/model/ami.py
jajreidy: Ditto for this.

src/pushsource/_impl/model/ami.py
jajreidy: Ditto for this.

src/pushsource/_impl/model/ami.py
ashwini3326: In this case I am removing this metadata for now, in case we need this one in future, we can add it.

ashwini3326:
based on the discussion above, I am removing the following metadata for now-
arch
image_name
virt_type
operating_system_version
operating_system_name
accounts
group

src/pushsource/_impl/model/ami.py
```diff
@@ -99,6 +99,20 @@ class AmiPushItem(VMIPushItem):
...
+    usage_instructions = attr.ib(type=str, default=None, validator=optional_str)
+    """Instructions for use of AMI"""
+
+    recommended_instance_type = attr.ib(type=str, default=None, validator=optional_str)
+    """Recommended instance type for AMI"""
+
+    product_type = attr.ib(type=str, default=None, validator=optional_str)
+    """Type of product e.g. "AMIProduct", "SaasProduct", "ServerProduct". """
     image_id = attr.ib(type=str, default=None, validator=optional_str)
     """AMI Image ID used to reference the image in AWS."""
```
JAVGan, line 117: Can you please add a newline to separate these attributes?

ashwini3326: done.

src/pushsource/_impl/model/ami.py
```diff
@@ -99,6 +99,21 @@ class AmiPushItem(VMIPushItem):
     )
     """Billing codes associated with this image."""
 
+    description = attr.ib(type=str, default=None, validator=optional_str)
+    """A description for the image"""
```
JAVGan, line 103: Sorry, I missed this one.

I believe we can remove it as well and use the [description](https://github.com/release-engineering/pushsource/blob/master/src/pushsource/_impl/model/vms.py#L57-L58) from the super class `VMIPushItem`.

@jajreidy WDYT?

jajreidy: Ah yeah that makes sense.

src/pushsource/_impl/model/ami.py
```diff
@@ -99,6 +99,21 @@ class AmiPushItem(VMIPushItem):
...
+
+    release_notes = attr.ib(type=str, default=None, validator=optional_str)
+    """"Release notes for AMI """
```
rohanpm, line 106: Could these doc strings please be a little more detailed?

What are the "release notes", is it like a changelog?  Roughly how long is it?  (e.g. a single sentence, a few sentences...)

src/pushsource/_impl/model/ami.py
```diff
@@ -99,6 +99,21 @@ class AmiPushItem(VMIPushItem):
...
+
+    usage_instructions = attr.ib(type=str, default=None, validator=optional_str)
+    """Instructions for use of AMI"""
```
rohanpm, line 109: Can this be a little more detailed too please?  Can you give an example of a string which would appear in this field?

src/pushsource/_impl/model/ami.py
```diff
@@ -99,6 +99,21 @@ class AmiPushItem(VMIPushItem):
...
+
+    recommended_instance_type = attr.ib(type=str, default=None, validator=optional_str)
+    """Recommended instance type for AMI"""
```
rohanpm, line 112: And this is also not entirely clear.

If this is the same thing as https://aws.amazon.com/ec2/instance-types/ , probably listing an example or two would be enough to make that clear, e.g. "m6in.8xlarge".

src/pushsource/_impl/model/ami.py
```diff
@@ -99,6 +99,21 @@ class AmiPushItem(VMIPushItem):
...
+
+    product_type = attr.ib(type=str, default=None, validator=optional_str)
+    """Type of product e.g. "AMIProduct", "SaasProduct", "ServerProduct". """
```
rohanpm, line 115: Where does this list of types come from?

jajreidy: At the moment, AWS doesn't have any list for these product types. I have asked them to include them in the docs in a meeting but they're not added ATM. So to get these I've just used the Marketplace webpage and watched the API calls for the differing product types.

rohanpm: So it's defined on the AWS side, not something we've made up?

There's got to be something we can link to or add to the doc here to disambiguate this.  Even if we don't know what all the valid values are, we could say something like "This field is known as (foo) when using (some particular AWS API/command)".

If you're not sure, could you tell me:

- How do we end up using the field?  Do we end up giving it as input to some AWS API?
- Could you tell me how, using either AWS CLI or AWS console, I can get AWS to show me one of these values.

jajreidy: product_type is used for the EntityType on function calls using Boto3. 
for describe an entity such as here:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/marketplace-catalog.html?highlight=awsmarketplace#MarketplaceCatalog.Client.describe_entity

or when creating a changeset like here:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/marketplace-catalog.html?highlight=awsmarketplace#MarketplaceCatalog.Client.start_change_set

You can look at the entities we already use by using list_entities here:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/marketplace-catalog.html?highlight=awsmarketplace#MarketplaceCatalog.Client.list_entities

You can also navigate to the Marketplace and look at the calls when navigating the Product drop down. They will look somewhat like this

https://aws.amazon.com/marketplace/management/products/api/v2/products?entityType=ServerProduct&maxResults=20&sortBy=LastModifiedDate&sortOrder=

There is the ability to use it with Offer but I couldn't find much on the usage of that. It's not great but it's all I've found so far.

src/pushsource/_impl/model/ami.py
ashwini3326: Yes, It is same as ec2 instance-type, I will update it with some references.

src/pushsource/_impl/model/ami.py
rohanpm: Thanks.  I wasn't able to find 'products' API calls myself, maybe we are getting different views of the marketplace, but in any case the usage as EntityType seems like enough info.

If I'm getting this right, then valid values for this field would be only whatever values for EntityType are accepted by the Marketplace Catalog API.

If so, I don't think we should give this a generic name like `product_type`.  Other code might be interested in a kind of "product type" which is unrelated to the marketplace catalog and which can't use the same values.  In this case, it seems like it'd be better to name the field more precisely around how it's used.  If values are specific to this one AWS API then it can't really be used for anything else.

So, may I suggest a better name for this would be something like `marketplace_entity_type` with a doc string mentioning something like "Entity type, as used by the AWS Marketplace Catalog API" (and with the examples retained).

jajreidy: That makes sense, I've been mulling over changing this to entity_type for awhile but marketplace_entity_type is more precise.

src/pushsource/_impl/model/ami.py
ashwini3326: updated this one with more detailed comment.

src/pushsource/_impl/model/ami.py
```diff
@@ -99,9 +99,42 @@ class AmiPushItem(VMIPushItem):
     )
     """Billing codes associated with this image."""
 
+    release_notes = attr.ib(type=str, default=None, validator=optional_str)
+    """"Notes regarding changes in the newest version. URLs are supported.
```
rohanpm, line 103: What does it mean for URLs to be "supported"?  If it's just a string, couldn't you put anything in it?

src/pushsource/_impl/model/ami.py
```diff
@@ -99,9 +99,42 @@ class AmiPushItem(VMIPushItem):
...
+    
+    SPECIFICATIONS:
```
rohanpm, line 105: In general, I think the docs should be consistent with other classes and fields, none of which use this kind of "SPECIFICATIONS" subheader.  The current doc style would be to write something like:

> If present, the value must not be blank and must not exceed 30,000 characters.

However, as these limits have now been documented, I wonder - would it be better to validate them instead of documenting them?  It'd be best to detect anything exceeding the limits as early as possible.

src/pushsource/_impl/model/ami.py
```diff
@@ -99,9 +99,42 @@ class AmiPushItem(VMIPushItem):
...
+    * Open text field
+    * 1 character MIN
+    * 30000 characters MAX
+    """
+
+    usage_instructions = attr.ib(type=str, default=None, validator=optional_str)
+    """Step by step instructions for the end-user to launch, configure and access the product. 
+    Write with a less-technical customer in mind. Can include URL to more comprehensive steps.
```
rohanpm, line 113: Please keep a blank line after the first sentence, like this:

```suggestion
    """Step by step instructions for the end-user to launch, configure and access the product. 

    Write with a less-technical customer in mind. Can include URL to more comprehensive steps.
```

The first sentence serves as a summary, then more details can appear in later paragraphs.

src/pushsource/_impl/model/ami.py
```diff
@@ -99,9 +99,42 @@ class AmiPushItem(VMIPushItem):
...
+    """Recommended instance type for AMI e.g. "t2.micro", "r6a.large" """
+
+    marketplace_entity_type = attr.ib(type=str, default=None, validator=optional_str)
+    """Type of entity e.g. "AMIProduct", "SaasProduct", "ServerProduct". """
+
     image_id = attr.ib(type=str, default=None, validator=optional_str)
     """AMI Image ID used to reference the image in AWS."""
 
+    scanning_port = attr.ib(type=str, default=None, validator=optional_str)
+    """AMI scanning port, used when importing the AMI into AWS Marketplace to validate the AMI."""
```
rohanpm, line 130: Could you give me some detail what this is please?  The reason is that it's declared as a string, but generally if we're talking about ports, `int` comes to mind.

rohanpm (changes requested):

src/pushsource/_impl/model/ami.py
ashwini3326: Yes, that's a good things, I will remove specifications.

src/pushsource/_impl/model/ami.py
ashwini3326: ok got it.

src/pushsource/_impl/model/ami.py
ashwini3326: updated this one.

src/pushsource/_impl/model/ami.py
ashwini3326: I have added more info in new commit.

src/pushsource/_impl/model/ami.py
ashwini3326: I did included the 'URL' part as it was mentioned in AWS info for release_notes, I think it means if they found any url they will be highlighting that, but in this case too, I think explicitly mentioning this one does not make it any better, so I will remove it.

src/pushsource/_impl/model/ami.py
ashwini3326: @jajreidy what do you think about this one?

ashwini3326: I just checked, that supposed to be int, my bad.

src/pushsource/_impl/model/ami.py
```diff
@@ -99,9 +99,33 @@ class AmiPushItem(VMIPushItem):
...
     """AMI Image ID used to reference the image in AWS."""
 
+    scanning_port = attr.ib(type=int, default=None, validator=instance_of(int))
+    """AMI scanning port, used when importing the AMI into AWS Marketplace to validate the AMI."""
+
+    user_name = attr.ib(type=str, default=None, validator=optional_str)
+    """The username used to login to the AMI. (Usually set to ec2-user)"""
+
+    security_groups = attr.ib(type=str, default=None, validator=optional_str)
+    """Automatically created security groups for the product. """
```
rohanpm, line 127: Group(s) (plural) but the type is a string?

If correct, how does the string contain more than one group?  e.g. comma-separated?  This detail should be mentioned.

tests/baseline/cases/staged-simple-ami-bc.yml
```diff
@@ -21,9 +21,11 @@ items:
     - dest1
     ena_support: true
     image_id: null
+    marketplace_entity_type: null
```
rohanpm, line 24: Ideally there should be some tests having non-null values for the new fields.  Otherwise it has never been tested that the fields can store any value other than None.

src/pushsource/_impl/model/ami.py
ashwini3326: fixed this one.

src/pushsource/_impl/model/ami.py
ashwini3326: I checked, it is supposed to be some list of strings, which could be converted to a meaningful JSON with some converter. I am applying the fix.

tests/baseline/cases/staged-simple-ami-bc.yml
ashwini3326: ok, I will populate the new fields of new metadata in the test files.

src/pushsource/_impl/model/ami.py
jajreidy: This will need to be something similar to BillingCode but it'll need to be a List[SecurityGroup] so that the data ends up looking something like this:

"SecurityGroups": [
                                {
                                    "FromPort": 22,
                                    "IpProtocol": "tcp",
                                    "IpRanges": [
                                        "[0.0.0.0/0](http://0.0.0.0/0)"
                                    ],
                                    "ToPort": 22
                                }
                            ]

src/pushsource/_impl/model/ami.py
rohanpm: Since the default is None, the validator has to allow for a value of None.  Right now it strictly requires int.

This is the reason for some tests failing with:

```
TypeError: ("'scanning_port' must be <class 'int'> (got None that is a <class 'NoneType'>)
```

JAVGan: @ashwini3326 in this case, you can use `optional` as validator:

```
    scanning_port = attr.ib(type=int, default=None, validator=optional(instance_of(int)))
    """AMI scanning port, used when importing the AMI into AWS Marketplace to validate the AMI."""
```

src/pushsource/_impl/model/ami.py
```diff
@@ -99,9 +99,33 @@ class AmiPushItem(VMIPushItem):
...
+
     @classmethod
     def _from_data(cls, data):
```
rohanpm, line 130: I see now that it's not possible to load any AMI push items right now using the new fields (at any non-default value), because:

- right now, the only supported way to load AMIs is by `staged` backend
- that loads items via this `_from_data` method
- but none of the new fields are copied across here

So there is no way to actually load/test AmiPushItem with the new fields as of today.

I think your end goal is to make this work with `errata` backend, but it seems all the new fields ought to be supported when using `staged` backend as well.

I suggest to add handling for all the new fields here in `_from_data`.  Then, if you were to add some example values for the new fields into tests/staged/data/simple_ami/staged.yaml, the fields would be covered by an existing test.  Alternatively the simple_ami test could be copied and modified as a new test.

tests/baseline/cases/staged-simple-ami-bc.yml
rohanpm: It seems like there are some code changes needed before this is possible.  Added a comment above on `_from_data` method.

src/pushsource/_impl/model/ami.py
ashwini3326: Implemented this one.

src/pushsource/_impl/model/ami.py
ashwini3326: Implemented.

ashwini3326:
@rohanpm @JAVGan @jajreidy  PTAL

JAVGan (approved):
LGTM

src/pushsource/_impl/model/ami.py
```diff
@@ -107,6 +160,18 @@ class AmiPushItem(VMIPushItem):
...
+    scanning_port = attr.ib(type=int, default=0, validator=optional(instance_of(int)))
+    """AMI scanning port, used when importing the AMI into AWS Marketplace to validate the AMI."""
+
+    user_name = attr.ib(type=str, default=None, validator=optional_str)
+    """The username used to login to the AMI. (Usually set to ec2-user)"""
+
+    version_title = attr.ib(type=str, default=None, validator=optional_str)
+    """The title given to a version. This will display in AWS Marketplace as the name of the version."""
+
+    security_groups = attr.ib(type=list, default=attr.Factory(frozenlist))
```
rohanpm, line 172: Please ensure there's `converter=frozenlist` as well so that the object remains immutable.

src/pushsource/_impl/model/ami.py
rohanpm: Is the new default of 0 valid on the AWS side?   How is it interpreted?  (Are there some docs about it?)

In general, 0 isn't a valid port number.  If that is the case here as well, it seems like the previous default of None was preferable.  Otherwise we're basically using two different values to indicate "no port number" which seems unnecessarily complex.

src/pushsource/_impl/schema/staged-schema.yaml
```diff
@@ -104,6 +104,53 @@ definitions:
...
+                $ref: "#/definitions/optional_string"
+
+            marketplace_entity_type:
+                $ref: "#/definitions/optional_string"
+
+            scanning_port:
+                type:
+                    - number
+                    - "null"
+                minimum: 0
```
rohanpm, line 123: Connecting to a comment above, what does 0 mean actually?

src/pushsource/_impl/schema/staged-schema.yaml
```diff
@@ -104,6 +104,53 @@ definitions:
...
+                    - array
+                    - "null"
+                items:
+                    type: object
+
+                properties:
+                    from_port:
+                        type: number
+                        minimum: 0
+                        maximum: 65353
```
rohanpm, line 141:
```suggestion
                        maximum: 65535
```

src/pushsource/_impl/schema/staged-schema.yaml
```diff
@@ -104,6 +104,53 @@ definitions:
...
+                    ip_protocol:
+                        type: string
+                    ip_ranges:
+                        type: array
+                        items:
+                            type: string
+                    to_port:
+                        type: number
+                        minimum: 0
+                        maximum: 65353
```
rohanpm, line 152:
```suggestion
                        maximum: 65535
```

src/pushsource/_impl/schema/staged-schema.yaml
```diff
@@ -104,6 +104,53 @@ definitions:
...
+            version_title:
+                $ref: "#/definitions/optional_string"
+            security_groups:
+                type:
+                    - array
+                    - "null"
+                items:
+                    type: object
+
+                properties:
```
rohanpm, line 137: I don't think this works, doesn't this part need to be indented by one more level so it's nested under "items"?  Otherwise the properties are being applied onto security_groups (the list), rather than the items within the list.

If in doubt - you can write a test which intentionally provides some invalid values under security_groups and check whether it really fails at these properties.

tests/baseline/cases/staged-simple-ami-bc.yml
```diff
@@ -21,10 +21,12 @@ items:
     - dest1
     ena_support: true
     image_id: null
+    marketplace_entity_type: AMIProduct
```
rohanpm, line 24: Right now there are two tests loading AMI items through 'staged' source.  You've updated both of them to use the new fields at non-default values.

Can we please instead have:
1. at least one test which specifies the new fields / loads them at non-default values, and
2. at least one test which does not specify any of the new fields (thus applying the default values)

All tools which exist today for generating the staging areas will generate them like (2) so it's important we don't break that case, and therefore important we test it.

Note, you can always add some more cases rather than trying to cram it all into the two existing cases.

rohanpm (changes requested):

tests/pub/data/100/images.json
```diff
@@ -32,6 +32,14 @@
     "state": "PUSHED", 
     "type": "hourly", 
     "virtualization": "hvm", 
-    "volume": "gp2"
+    "volume": "gp2",
```
rohanpm, line 35: Similarly to the comments about the staging area testdata:

100% of pub tasks which exist today are not going to have these new fields in images.json.

So, can we please keep at least one test which covers the case where all of those fields are absent?

src/pushsource/_impl/model/ami.py
rohanpm: Ashwini confirmed that 0 is accepted by the AWS APIs, so I think this is OK.  `None` means "we don't have a value for the field", 0 means whatever the AWS API does with 0.

src/pushsource/_impl/schema/staged-schema.yaml
rohanpm: Actually, it does not really matter what this means, only if it matches what the AWS API accepts.  Accepting 0 is correct if AWS also accepts 0.

src/pushsource/_impl/model/ami.py
```diff
@@ -55,6 +55,44 @@ def _from_data(cls, data):
...
+    """The IP protocol name ( tcp , udp , icmp , icmpv6 )."""
+
+    ip_ranges = attr.ib(type=list, converter=frozenlist)
+    """List of the IPv4 ranges, for example ['22.22.22.22', '33.33.33.33'].
+
+    :type: list[str]
+    """
+
+    to_port = attr.ib(type=int, validator=instance_of(int))
+    """If the protocol is TCP or UDP, this is the end of the port range.
```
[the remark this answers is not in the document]
ashwini3326: updated the to_port docs to say 'it specify the codes'

src/pushsource/_impl/schema/staged-schema.yaml
ashwini3326: As it accepts 0 so I am leaving this unchanged.

src/pushsource/_impl/schema/staged-schema.yaml
ashwini3326: did fixed this one to validate the security groups data properly

tests/baseline/cases/staged-simple-ami-bc.yml
ashwini3326: I have removed newly added metadata from simple-ami

tests/pub/data/100/images.json
ashwini3326: I have removed newly added metadata from 100/images.json

ashwini3326:
@rohanpm PTAL whenever you have time.

rohanpm (approved):

rohanpm:
Looks good, should it be released immediately or anyone else want to review one more time?

JAVGan:
> Looks good, should it be released immediately or anyone else want to review one more time?

Feel free to release it 🙂

PR #305, opened 2023-02-07
Added container_list attribute to Erratum model

midnightercz:


src/pushsource/_impl/schema/errata-schema.yaml
```diff
@@ -149,6 +155,8 @@ properties:
         $ref: "#/definitions/reference_list"
     pkglist:
         $ref: "#/definitions/pkg_collection_list"
+    containerlist:
+        $ref: "#/definitions/container_lits"
```
rohanpm, line 159:
```suggestion
        $ref: "#/definitions/container_list"
```

src/pushsource/_impl/schema/errata-schema.yaml
```diff
@@ -149,6 +155,8 @@ properties:
         $ref: "#/definitions/reference_list"
     pkglist:
         $ref: "#/definitions/pkg_collection_list"
+    containerlist:
```
rohanpm, line 158: For the schema to work correctly, the property name has to match the name you're looking for in `_from_data` method.  Right now the schema is going to check a property named `containerlist` but then the code ends up using a field named `container_list`, so it's not actually being checked by the schema.

So either this should be `container_list` or from_data should be looking for `containerlist`.

src/pushsource/_impl/model/erratum.py
```diff
@@ -265,6 +265,14 @@ class ErratumPushItem(PushItem):
     :type: list[ErratumPackageCollection]
     """
 
+    container_list = attr.ib(
+        type=list, default=attr.Factory(frozenlist), converter=frozenlist
+    )
+    """A list of containers ssociated with the advisory.
```
rohanpm, line 271:
```suggestion
    """A list of container images associated with the advisory.
```

Also, could you please be a little more specific here?  Are they image pull specs?  By digest, by tag or both?

Also, does the order of this list matter?  If it doesn't, can we please canonicalize the data by sorting the list - this will ensure that an erratum with container_list of [A, B] vs a container_list of [B, A] are considered to be the same object.

tests/baseline/cases/errata-containers.yml
```diff
@@ -1779,6 +1779,7 @@ items:
 - ErratumPushItem:
```
rohanpm, line 1779: Could you please ensure there is at least one test which results in a non-empty container_list?

rohanpm (changes requested):

src/pushsource/_impl/model/erratum.py
```diff
@@ -368,12 +376,20 @@ def _from_data(cls, data):
         if data.get("cdn_repo"):
             kwargs["dest"] = data["cdn_repo"]
 
+        # If source of data is ET/staging directory, try to get container_list directly
+        if data.get("container_list"):
+            kwargs["container_list"] = data["container_list"]
```
rohanpm, line 381: This doesn't seem right, because the  model is a flat list of strings, but this is not what ET returns.

An example of what ET returns can be found at https://url.corp.redhat.com/26b7928

rohanpm (changes requested):
The base of this pull request is about 2 months old, can you please rebase it?  There were some important changes meanwhile such as dropping py2 support.

src/pushsource/_impl/model/erratum.py
midnightercz: fixed

tests/baseline/cases/errata-containers.yml
midnightercz: added

src/pushsource/_impl/schema/errata-schema.yaml
midnightercz: fixed

tests/baseline/cases/errata-containers-legacy-repos.yml
```diff
@@ -1779,6 +1779,16 @@ items:
 - ErratumPushItem:
     build: null
     build_info: null
+    container_list:
+    - openshift4/ose-local-storage-rhel7-operator-metadata:
+        digest: sha256:2321a7d13d9fa53f05437663cf2dc217d15f3cda4b67076c941b10f0491cf9d7
+        images:
+          docker-image-sha256:05649a19718fde131372c761d359302ccbe81f9744d2893ac4c826b05d670206.ppc64le.tar.gz:
```
rohanpm, line 1786: I'm questioning whether it really makes sense to expose the koji archive names here.

I know that's what ET is doing, but you already wrote code to replace the koji *build* NVR (like local-storage-operator-metadata-container-v4.3.28.202006290519.p0.prod-1) with the more useful and generic repo names (like openshift4/ose-local-storage-rhel7-operator-metadata).  So you already set the precedent that we'll patch up the data provided by ET into something more useful for the use-case.

In that case, does it make more sense to expose the arch here instead?  I think the users of this data wanted to know the arch of each image, didn't they?  And right now they'd probably be forced to write some code like: split the key on "." and the third-last element is the image's arch.

Maybe we should take care of that for them?  Right now the format seems like a quirk of how ET and koji work together, and we can take care of it right here where we query ET since we know how ET works.  We can then expose the data as one of these formats:

```yaml
images:
- arch: s390x
  digest: (...the digest)
- arch: x86_64
  digest: (...the digest)
```

...or:

```yaml
images:
  s390x:
    digest: (...the digest)
  x86_64:
    digest: (...the digest)
```

Either one I suppose is more useful for the code wanting to access this data.

It would also be easier to ensure that data remains compatible over time, seeing as there is no guarantee that container images are ever stored into separate files per arch like this - that's just a detail of how it currently works in koji.

rohanpm (changes requested):

tests/baseline/cases/errata-containers-legacy-repos.yml
midnightercz: I was also thinking the same way. But same as the format of storing images can change, so does the archive name. And as pub is only the middle man here I didn't want  to involve here more transformations than necessary.

rohanpm (changes requested):
Sorry, I prepared these comments yesterday but failed to submit them.

src/pushsource/_impl/backend/errata_source/errata_source.py
```diff
@@ -151,7 +151,27 @@ def _advisory_ids(self):
...
+            for clistitem in raw_metadata["container_list"][:]:
+                new_clistitem = {}
+                for build in list(clistitem.keys()):
+                    build_repos = raw.advisory_cdn_docker_file_list[build]["docker"][
+                        "target"
+                    ]["external_repos"]
+                    val = clistitem[build]
+                    new_val = {"digest": val["digest"], "images": {}}
+                    for image, digest_dict in val["images"].items():
+                        arch = image.split(".")[-3]
```
rohanpm, line 167: Minor: a comment or two here explaining what it's doing would be valuable (e.g. an example value of `image`).

src/pushsource/_impl/model/erratum.py
```diff
@@ -265,6 +266,34 @@ class ErratumPushItem(PushItem):
     :type: list[ErratumPackageCollection]
     """
 
+    container_list = attr.ib(
+        type=list, default=attr.Factory(frozenlist), converter=frozenlist
+    )
+    """A list of container images associated with the advisory.
+
+    Example of the list:
```
rohanpm, line 274: This won't render correctly in the docs, you'll have to do something like:

- blank space after this line
- put the example in an indented `.. code-block:: python`

You can find examples if you grep for `code-block::`.

src/pushsource/_impl/model/erratum.py
```diff
@@ -265,6 +266,34 @@ class ErratumPushItem(PushItem):
...
+    "container_list": [
+        {
+            "openshift4/ose-local-storage-rhel7-operator-metadata": {
+                "digest": "sha256:2321a7d13d9fa53f05437663cf2dc217d15f3cda4b67076c941b10f0491cf9d7",
+                "images": {
+                    "docker-image-sha256:06d1c5e4fa6a5d1ff868388f3feadf193d04128b62d1181e37fe4ab8ecda27e1.s390x.tar.gz": {
```
rohanpm, line 280: This is not correct, e.g. this should just be s390x.

tests/staged/data/simple_errata/dest1/ERRATA/advisory2.json
```diff
@@ -159,6 +159,24 @@
   "pulp_user_metadata": {
     "content_types": [
       "rpm"
+    ],
+    "container_list": [
```
rohanpm, line 163: The sample data here doesn't seem to be correct as the format is neither the same as what ET produces (as build NVRs were replaced with a repo name), nor what ErratumPushItem.container_list is expected to be (as image archive names have not been replaced with arches).

What format is supposed to be used for staged errata?  It seems that it should be either the ET native format or the ErratumPushItem format and not something in-between.

In either case we should arrange it so that the keys under `images` are expected to be arch names in the staged case, same as it is in the ET case.

rohanpm (changes requested):

tests/staged/data/simple_errata/dest1/ERRATA/advisory2.json
midnightercz: My understanding was that it should be provided in already transformed format. So images needs to be replaced with architectures

src/pushsource/_impl/model/erratum.py
```diff
@@ -265,6 +266,35 @@ class ErratumPushItem(PushItem):
...
     """
 
+    container_list = attr.ib(
+        type=list, default=attr.Factory(frozenlist), converter=frozenlist
+    )
+    """A list of container images associated with the advisory.
+
+    Example of the list:
+    
+    .. code-block:: python
```
rohanpm, line 276: You need a blank line after the code-block as well.  It should be possible to test this by running `tox -e docs` and checking the generated docs.

```suggestion
    .. code-block:: python

```

rohanpm (changes requested):

rohanpm (approved):

PR #306, opened 2023-02-08
Don't log during stagedir wait if wait is disabled

rohanpm:
The addition of this log caused some log-based tests in other projects to start failing due to new log messages appearing like this:

      [    INFO] Directory '/tmp/aws_staged' has mtime 2023-02-05T20:28:28+00:00,
      being more than 0 seconds old. Skipping the wait.

I don't think the log message makes sense anyway if the waiting feature has been disabled by setting the threshold to 0 seconds, so let's not log in that case.

### DIFF
--- a/src/pushsource/_impl/backend/staged/staged_source.py
+++ b/src/pushsource/_impl/backend/staged/staged_source.py
@@ -164,7 +164,7 @@ class StagedSource(
                 wait_time,
             )
             time.sleep(wait_time)
-        else:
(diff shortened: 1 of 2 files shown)


rohanpm:
Needed to fix https://github.com/release-engineering/pubtools-ami/pull/86.

rbikar (approved):

rajulkumar (approved):

PR #311, opened 2023-02-27
Fixed frozendict convertor to core.frozendict

midnightercz:
validation of container_list in erratum model is set to check if instance of frozen dict is frozendict.core.frozendict which is not "compatible" with frozendict.frozendict

### DIFF
--- a/src/pushsource/_impl/model/container.py
+++ b/src/pushsource/_impl/model/container.py
@@ -1,10 +1,10 @@
 import re
 
 from frozenlist2 import frozenlist
+from frozendict.core import frozendict  # pylint: disable=no-name-in-module
 
 from .base import PushItem
 from .. import compat_attr as attr
-from frozendict import frozendict
 from .conv import instance_of, optional_str
 
 
--- a/src/pushsource/_impl/model/conv.py
+++ b/src/pushsource/_impl/model/conv.py
@@ -5,7 +5,7 @@ from functools import partial
 import re
 
 from frozenlist2 import frozenlist
-from frozendict.core import frozendict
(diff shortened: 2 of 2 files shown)


rohanpm (changes requested):
This causes tests to fail.

Can you explain a bit what this is trying to fix?  The container_list field in this library doesn't even have a validator:

```
    container_list = attr.ib(
        type=list, default=attr.Factory(frozenlist), converter=frozenlist
    )
```

Is it a validator in pubtools-pulplib you're referring to?

If so, I'd say the bug is in pubtools-pulplib as it should be capable of converting the incoming data from any kind of Mapping, it should not be strictly requiring a specific implementation of a dict.

midnightercz:
> Is it a validator in pubtools-pulplib you're referring to?
Yeah, it't validator in pubtools-pulplib I'm referring to. 

> If so, I'd say the bug is in pubtools-pulplib as it should be capable of converting the incoming data from any kind of Mapping, it should not be strictly requiring a specific implementation of a dict.
It is able to convert anything if it's not already frozen. And it looks like frozendict.core.frozendict is subclass if frozendict.frozendict. Validator then however complains because it was checking if if situation is vice-versa. So easiest way to fix this seems ot unify frozendict imports to frozendict.core

rohanpm (approved):
Hmm, I think that we probably made a mistake starting to use frozendict.core.  
Looking at the frozendict project, 'core' module is not a part of the documented API and it even ends up being deleted from the frozendict module in some cases.

However, it looks like that usage was already introduced in April 2022 prior to the recent changes, so I guess that's a problem for another day.

PR #315, opened 2023-03-06
Add public_image flag for AMI images [RHELDST-16452]

MichalHaluza:
The `public_image` flag designates whether the AMI image is allowed to be released publicly (shared with group "all") as some images are expected to be "shipped live" by going through all the standard steps with the exception of being shared publicly.

This commit also fixes logic for getting the `ena_support` flag. Previously, when the flag was set to `False`, the
`AmiPushItem._from_data()` method unintentionally converted it to `None`. Now it remains set to `False`.

This PR relates to https://github.com/release-engineering/pubtools-ami/pull/93

### DIFF
--- a/src/pushsource/_impl/model/ami.py
+++ b/src/pushsource/_impl/model/ami.py
@@ -102,6 +102,11 @@ class AmiPushItem(VMIPushItem):
     image_id = attr.ib(type=str, default=None, validator=optional_str)
     """AMI Image ID used to reference the image in AWS."""
 
+    public_image = attr.ib(
+        type=bool, default=None, validator=optional(instance_of(bool))
+    )
+    """``True`` if the image is allowed to be released publicly (shared with group "all")."""
+
     @classmethod
     def _from_data(cls, data):
         """Instantiate AmiPushItem from raw list or dict"""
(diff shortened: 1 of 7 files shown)


rohanpm (approved):

src/pushsource/_impl/model/ami.py
```diff
@@ -102,6 +102,11 @@ class AmiPushItem(VMIPushItem):
     image_id = attr.ib(type=str, default=None, validator=optional_str)
     """AMI Image ID used to reference the image in AWS."""
 
+    public_image = attr.ib(
+        type=bool, default=None, validator=optional(instance_of(bool))
```
rohanpm, line 106: Connecting with a discussion from JIRA:

The old behavior from pubtools-ami was to treat an image as public if type == "hourly".  If the default is now None, and that's interpreted as False, that can be considered a backwards-incompatible change which will make older staging areas not work as expected.

One way to fix this would be to derive the default here from `type == "hourly"`, or on the `_from_data` method below.

Another way would be leaving the default as None to accurately indicate "unknown" and then having pubtools-ami apply backwards-compatible default behavior if the value is None (rather than treating it as False).

MichalHaluza: I'll opt for the pubtools-ami way. The current behavior there already contains the check for hourly images, so the new behavior will only add to that check. Something in the lines of
```python
public_image = push_item.public_image
if public_image is None and image_type == "hourly":
    public_image = True
```

PR #316, opened 2023-03-06
Address warnings from pylint and sphinx

MichalHaluza:
Most notably, this commit adds the docs for recently added pub source into index. The rest of the changes are to keep pylint happy and don't modify the functionality.

### DIFF
--- a/src/pushsource/_impl/model/vms.py
+++ b/src/pushsource/_impl/model/vms.py
@@ -61,7 +61,9 @@ class VMIPushItem(PushItem):
     def __validate_release(self, attribute, value):  # pylint: disable=unused-argument
         # Strict (unidiomatic) type check ensures that subclasses
         # of VMIRelease can not be associated with VHDPushItem
-        if value and type(value) is not self._RELEASE_TYPE:  # pylint: disable=unidiomatic-typecheck
+        if (
+            value and type(value) is not self._RELEASE_TYPE
+        ):  # pylint: disable=unidiomatic-typecheck
             raise ValueError(
                 'The release type must be "%s"' % self._RELEASE_TYPE.__name__
             )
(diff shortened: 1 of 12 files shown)


MichalHaluza:
Looks like the missing docs are addressed by https://github.com/release-engineering/pushsource/pull/313 already. This PR should be rebased once 313 is merged.

src/pushsource/_impl/model/vms.py
```diff
@@ -58,9 +58,10 @@ class VMIPushItem(PushItem):
     """A brief human-readable description of the image."""
 
     @release.validator
-    def __validate_release(self, attribute, value):
-        # pylint: disable=unused-argument
-        if value and not type(value) == self._RELEASE_TYPE:
+    def __validate_release(self, attribute, value):  # pylint: disable=unused-argument
+        if (
+            value and type(value) is not self._RELEASE_TYPE
```
rohanpm, line 63: I believe it could actually be made idiomatic, I don't know a specific reason why we need to enforce a precise release type rather than `isinstance`, and there are some downsides to the precise check - e.g. I don't know a reason why a test should be blocked from providing its own subclass with extra testing features.

I noticed it as a minor issue during code review of this recently-introduced code.  I may have forgotten to mention it or didn't bother since that review was already taking a long time.

rohanpm (approved):

src/pushsource/_impl/model/vms.py
MichalHaluza: I did try to use `isinstance()` at first, but that made the [test_vmi_release()](https://github.com/release-engineering/pushsource/blob/09dc42211195d153c85624845e15b7c4b4023b9f/tests/model/test_vhd.py#L39-L68) fail as the `AmiRelease`, which is a subclass of the `VMIRelease` is not expected to be associated with `VHDPushItem`. The docstring "_Ensure that subclasses of VMIRelease can not be associated with VHDPushItem_" suggests that unidiomatic strict type check is the expected one as well.

rbikar (approved):

PR #331, opened 2023-04-18
Adds ability to create push items for VMIs from Errata

jajreidy:
Allows Errata to be a source when uploading images to clouds. @JAVGan @lslebodn @ashwini3326 PTAL

### DIFF
--- a/docs/sources/errata.rst
+++ b/docs/sources/errata.rst
@@ -9,6 +9,7 @@ Supported content types:
 * RPMs
 * Container images
 * Advisories
+* Virtual machine images
 
 
 errata source URLs
--- a/src/pushsource/_impl/backend/errata_source/errata_source.py
+++ b/src/pushsource/_impl/backend/errata_source/errata_source.py
@@ -193,6 +193,12 @@ class ErrataSource(Source):
             erratum, raw.advisory_cdn_docker_file_list
         )
 
+        appliance_image_list = raw.advisory_cdn_metadata.get("appliance_image_list")
+
+        items = items + self._push_items_from_appliance_image_list(
+            erratum, appliance_image_list
+        )
+
         return [erratum] + items
 
     def _push_items_from_container_manifests(self, erratum, docker_file_list):
@@ -244,6 +250,20 @@ class ErrataSource(Source):
 
                 item = attr.evolve(item, origin=erratum.name)
                 out.append(item)
+
+        return out
+
+    def _push_items_from_appliance_image_list(self, erratum, appliance_image_list):
+        out = []
+        if appliance_image_list:
+            nvr_list = []
+            for image in appliance_image_list:
+                nvr_list.extend([nvr for nvr in image.keys()])
+
+            with self._koji_source(vmi_build=nvr_list) as koji_source:
+                for push_item in koji_source:
+                    push_item = attr.evolve(push_item, origin=erratum.name)
+                    out.append(push_item)
 
         return out
 
--- a/tests/errata/test_errata_vmis.py
+++ b/tests/errata/test_errata_vmis.py
@@ -1,0 +1,220 @@
+import os
+import datetime
+from mock import patch
+
+import pytest
+
+from pushsource import (
+    Source,
+    VMIPushItem,
+    VMIRelease,
+    AmiPushItem,
+    AmiRelease,
+    VHDPushItem,
+    KojiBuildInfo,
+)
+
+
+def test_errata_ami_via_koji(fake_errata_tool, fake_koji, koji_dir):
+    """Errata source containing a module yields modules & RPMs taken
+    from koji source"""
+
+    source = Source.get(
+        "errata:https://errata.example.com",
+        errata="RHSA-2022:78146",
+        koji_source="koji:https://koji.example.com?basedir=%s" % koji_dir,
+    )
+
+    nvr = "rhel-ec2-8.7-1"
+    name = "rhel"
+    version = "8.7"
+    release = "1"
+
+    # Insert some data
+    fake_koji.build_data[1234] = {
+        "id": 1234,
+        "name": name,
+        "version": version,
+        "release": release,
+        "nvr": nvr,
+        "completion_time": "2022-12-20 12:30:25",
+        "extra": {"typeinfo": {"image": {}}},
+    }
+    fake_koji.build_data[nvr] = fake_koji.build_data[1234]
+
+    imgs = {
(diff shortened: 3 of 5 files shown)


docs/sources/errata.rst
```diff
@@ -9,6 +9,7 @@ Supported content types:
 * RPMs
 * Container images
 * Advisories
+* VMIs
```
JAVGan, line 12: Can you please change it to `Virtual machine images`?

Just to be consistent with [KojiSource](https://github.com/release-engineering/pushsource/blob/master/docs/sources/koji.rst)

```suggestion
* Virtual machine images
```

src/pushsource/_impl/backend/errata_source/errata_source.py
```diff
@@ -247,6 +251,22 @@ def _push_items_from_container_manifests(self, erratum, docker_file_list):
...
 
+    def _push_items_from_appliance_image_list(self, appliance_image_list):
+        if not appliance_image_list:
+            return []
+
+        nvr_list = []
+        out = []
+        for image in appliance_image_list:
+            for nvr in image.keys():
+                nvr_list.append(nvr)
```
JAVGan, line 262: Just a suggestion to simplify this:

```suggestion
        for image in appliance_image_list:
            nvr_list.extend([nvr for nvr in image.keys()])
```

src/pushsource/_impl/backend/errata_source/errata_source.py
```diff
@@ -247,6 +251,22 @@ def _push_items_from_container_manifests(self, erratum, docker_file_list):
 
         return out
 
+    def _push_items_from_appliance_image_list(self, appliance_image_list):
+        if not appliance_image_list:
+            return []
```
JAVGan, line 256: Another way around is to simply return `out` (which should be an empty list when not appliance_image_list)

```
        out = []
        
        if appliance_image_list:
            ...  # Do the for loop and fetch the push_items

        return out
```

JAVGan (commented):
Just minor comments

src/pushsource/_impl/backend/errata_source/errata_source.py
jajreidy: updated

src/pushsource/_impl/backend/errata_source/errata_source.py
jajreidy: updated

docs/sources/errata.rst
jajreidy: updated

JAVGan (approved):
LGTM

tests/errata/test_errata_vmis.py
```diff
@@ -0,0 +1,118 @@
...
+
+    fake_koji.insert_archives(archives=archives, build_nvr=nvr)
+
+    with source:
+        items = list(source)
+
+    vmi_items = [i for i in items if isinstance(i, VMIPushItem)]
+
+    for push_item in vmi_items:
+        assert isinstance(push_item, AmiPushItem)
```
rohanpm, line 62: It seems like at least one of these tests should be doing a bit more in-depth test of the generated push items, rather than just `isinstance`.

This test would pass even if the errata source generated AmiPushItem instances with nonsense in all the fields, wouldn't it?

src/pushsource/_impl/model/erratum.py
```diff
@@ -354,6 +354,14 @@ class ErratumPushItem(PushItem):
     :type: list[str]
     """
 
+    appliance_image_list = attr.ib(
```
rohanpm, line 357: I'm not convinced that this should be added as a public field on ErratumPushItem.

Do you have any plan to make use of it?  If so, can you share some details about it?  If no, please leave it out.  The info is anyway implied by the set of items you get back along with this item.

The problems I have with this field as currently defined (being a list of koji NVR):

(1) API shouldn't be unnecessarily tied to koji.

ErratumPushItem and VMIPushItem can already be generated without involving ET and koji at all (using a staging directory).  What would be put in this field in that case?  If it would be left blank - that means consumers have to be prepared to deal with a subtle difference between ErratumPushItem loaded from ET, and loaded from a staging dir, and they probably won't be aware of that.

(2) It's not consistent with or useful in the same way the other similar fields (pkglist for RPMs, and container_list for images) are useful.

Those other fields expose data which ends up going into customer portal and which makes sense to the end-user.  e.g. the names and checksums of RPMs makes sense, and the pull specs and manifest digests of container images make sense.  The NVRs of builds in koji don't make sense since that's an internal system the customers aren't aware of, don't have access to, and potentially could be replaced with something else.

TL;DR if you don't have an actual need for this, please don't add it; and if you *do* have a need for it then we need to discuss further what the format should be, because a list of koji NVRs is probably not good enough.

rohanpm (changes requested):

src/pushsource/_impl/model/erratum.py
jajreidy: I thought it was needed but it's not so I removed it.

tests/errata/test_errata_vmis.py
jajreidy: Added more checks to it, was able to see I wasn't setting origin correctly. So it's been updated.

rohanpm (approved):

rohanpm:
CI is having a problem with rpm-py-installer.  This is a cursed dependency which causes failures from time to time.  I'll have a look at it.

rohanpm:
> CI is having a problem with rpm-py-installer. This is a cursed dependency which causes failures from time to time. I'll have a look at it.

Looks like we're going to need https://github.com/release-engineering/pushsource/pull/332 submitted first, and then this rebased.

tests/errata/test_errata_vmis.py
```diff
@@ -0,0 +1,220 @@
...
+                arch="x86_64",
+                respin=1,
+                version="8.7",
+                base_product=None,
+                base_version=None,
+                variant=None,
+                type=None,
+            ),
+            description="",
+            type=None,
```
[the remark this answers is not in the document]
JAVGan: Yes, it's being defined in StArMap for our pushes

rohanpm:
CI is supposed to be fixed now.  Could you please rebase this change so we can get a clean CI run?

jajreidy:
> CI is supposed to be fixed now. Could you please rebase this change so we can get a clean CI run?

Done

PR #336, opened 2023-05-08
Update pull spec with unique tag [CLOUDDST-18636]

emilyzheng:
Replace the floating tag in pull spec with a unique tag.

### DIFF
--- a/src/pushsource/_impl/backend/koji_containers.py
+++ b/src/pushsource/_impl/backend/koji_containers.py
@@ -93,8 +93,9 @@ class ContainerArchiveHelper(object):
         ################## List pull specs - from build #################################
         build_raw_specs = self.build_index.get("pull") or {}
         build_digests = self.build_index.get("digests") or {}
+        unique_tags = self.build_index.get("unique_tags") or []
         list_digest_specs = get_digest_specs(build_raw_specs, build_digests)
-        list_tag_specs = get_tag_specs(build_raw_specs)
+        list_tag_specs = get_tag_specs(build_raw_specs, unique_tags)
 
         # Metadata does not explicitly tell us the media type per tag, but these tag
         # specs are known to be for manifest lists. We'll sanity check that this type
@@ -114,8 +115,16 @@ class ContainerArchiveHelper(object):
         )
 
 
-def get_tag_specs(raw_specs):
-    return [ContainerImagePullSpec._from_str(s) for s in raw_specs if "@" not in s]
+def get_tag_specs(raw_specs, unique_tags=None):
+    out = []
+    for raw_spec in raw_specs:
+        if "@" not in raw_spec:
+            # Replace the floating tag in spec with a unique tag
+            if unique_tags:
+                raw_spec = "{0}:{1}".format(raw_spec.split(":")[0], unique_tags[0])
+            spec = ContainerImagePullSpec._from_str(raw_spec)
+            out.append(spec)
+    return out
 
 
 def get_digest_specs(raw_specs, digests_map):
--- a/tests/baseline/cases/errata-containers-legacy-repos.yml
+++ b/tests/baseline/cases/errata-containers-legacy-repos.yml
@@ -46,7 +46,7 @@ items:
         - application/vnd.docker.distribution.manifest.list.v2+json
         registry: registry-proxy.engineering.redhat.com
         repository: rh-osbs/openshift-ose-cluster-logging-operator-metadata
-        tag: v4.3.28.202006290519.p0.prod-1
+        tag: rhaos-4.3-rhel-7-candidate-56520-20200707074906
       - media_types: []
         registry: registry-proxy.engineering.redhat.com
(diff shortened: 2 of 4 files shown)


rohanpm:
I think there's a couple of issues with this:

1. If the user of this info wants to ensure they're pulling the right thing, why don't they use digest_specs rather than tag_specs?

The docs at https://release-engineering.github.io/pushsource/model/containers.html#pushsource.ContainerImagePullInfo.tag_specs even warn about this:

> Note that, as all tags are mutable, it’s possible for this information to be outdated.
> It is recommended to pull by digest or at least cross-reference with digests.

So we seem to have some client code which ignores the suggestion to pull by digest and instead used mutable tags as if they were immutable.  And now to work around that bug, rather than doing as suggested, we're changing the behavior of `tag_specs` field to support the bad client.  Can we fix the client instead?

2. While *adding* unique_tags into tag_specs makes sense to me, it doesn't seem right that we would also be *removing* the floating tags.

Can we expose both?  `tag_specs` has always been a list.  I think the most reasonable interpretation is that it would contain all tags which could be used to access the image.  Not a subset which has been filtered to meet the needs of one particular client (and potentially doesn't meet the needs of some other client).

Put another way, if you look at the test data for e.g. openshift-ose-cluster-logging-operator-metadata , the v4.3.28.202006290519.p0.prod-1 tag is now disappearing from the output.  How do we know that nobody wants to see that tag?  I don't think we can safely conclude that.

So while making use of unique_tags seems like a good idea, I'd say these should be added to `tag_specs` and shouldn't replace anything which is currently there.

The docs also say that if you don't care about which tag you use, you can use `item.pull_info.tag_specs[0]`.  If unique tags are the best tags then (least likely to have changed), it would make sense that unique_tags are added to the *beginning* of `tag_specs`.

rohanpm (changes requested):
I think the approach is not quite right per comment https://github.com/release-engineering/pushsource/pull/336#issuecomment-1539193429 .

emilyzheng:
> I think the approach is not quite right per comment [#336 (comment)](https://github.com/release-engineering/pushsource/pull/336#issuecomment-1539193429) .

Thanks for the comments.

For 1, tags are needed for pulp repo sync (tags in [importer](https://docs.pulpproject.org/en/2.17/plugins/pulp_docker/tech-reference/importer.html)). Making pulp support repo sync from digest wouldn't be worth the effort since we are migrating out of pulp.

For 2, `contain all tags which could be used to access the image` wouldn't be fulfilled unless both floating_tags and unique_tags are added to tag_specs. Is it necessary to include all tags?

Pub uses only tag_specs[-1] and tag_specs[0], this change replaces the tag in tag_specs[0]. Any other clients I should consider?

rohanpm:
> For 1, tags are needed for pulp repo sync (tags in [importer](https://docs.pulpproject.org/en/2.17/plugins/pulp_docker/tech-reference/importer.html)). Making pulp support repo sync from digest wouldn't be worth the effort since we are migrating out of pulp.
> 

Right, that makes sense.

> For 2, `contain all tags which could be used to access the image` wouldn't be fulfilled unless both floating_tags and unique_tags are added to tag_specs.Is it necessary to include all tags?
> 
> Pub uses only tag_specs[-1] and tag_specs[0], this change replaces the tag in tag_specs[0]. Any other clients I should consider?

Well, we shouldn't be thinking only of Pub or the clients we know about.  We need to have behavior which makes sense in general and not just what fixes a bug in Pub.

However, having looked at this more broadly now I changed my mind, and I think the change as implemented is OK.

I see the info we get from koji is like this:

```yaml
        floating_tags:
        - latest
        - v4.3.28.202006290519.p0.prod
        pull:
        - registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-sriov-network-operator-metadata@sha256:16a04fde191ebd8858fa7f97605a9783be570e37d329eec599cfecf03b5cd8f2
        - registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-sriov-network-operator-metadata:v4.3.28.202006290519.p0.prod-1
        tags:
        - v4.3.28.202006290519.p0.prod-1
        unique_tags:
        - rhaos-4.3-rhel-7-candidate-88311-20200707073051
```

There are three different types of tag.  The ones we returned prior to this change was whatever was found in 'pull', which seems to match 'tags', which https://github.com/containerbuildsystem/atomic-reactor/blob/master/docs/koji.md says is "Primary tags" but there is no explanation of what that means.  I thought that the tags we were producing here would have been floating tags, but instead there's a separate `floating_tags` which has other values not found in `pull`...

So given that, today, the tags we're already producing are kind of arbitrary and don't seem to have a well-defined meaning, I guess it's not a problem to change them.

Personally, I would be worried that changing from tags like "v4.3.28.202006290519.p0.prod-1" to "rhaos-4.3-rhel-7-candidate-88311-20200707073051" during pulp sync might have unexpected side-effects, but I suppose you'll ensure it's tested.

rohanpm (approved):

querti (approved):

PR #337, opened 2023-05-09
Use rpmdyn in CI rather than rpm-py-installer

rohanpm:
Due to junaruga/rpm-py-installer#276, it was necessary to pin a particular version of virtualenv.

Let's undo that pin and try an alternative provider of RPM bindings instead. This library works a different way which does not require downloading and compiling from RPM sources, and is compatible with the newest versions of pip and other tools.

This is still somewhat experimental. One of the reasons for picking this project is that docs building here is still broken by rpm-py-installer, so the online docs have been broken for a little while; this should fix it.


rbikar:
Nice :) I guess I'll give `rpmdyn` a shot in ubi related proejects.

rbikar (approved):

PR #340, opened 2023-05-19
Allow KojiSource to store the build id

JAVGan:
This PR changes the `KojiSource` to store the build id for containers and VMIs whenever it's available.
    
It adds an optional attribute `id` into the `KojiBuildInfo` model to  support the change and update the related tests.
    
Refers to EXDSP-2099

### DIFF
--- a/src/pushsource/_impl/model/base.py
+++ b/src/pushsource/_impl/model/base.py
@@ -7,11 +7,13 @@ from frozenlist2 import frozenlist
 from .. import compat_attr as attr
 from .cache import TinyCache
 from .conv import (
+    convert_maybe,
     md5str,
     sha256str,
     upper_if_str,
     instance_of_str,
     instance_of,
+    optional,
     optional_str,
 )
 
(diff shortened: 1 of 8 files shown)


JAVGan:
@rohanpm @lslebodn @jajreidy PTAL

jajreidy:
LGTM

src/pushsource/_impl/model/base.py
```diff
@@ -42,6 +43,9 @@ class KojiBuildInfo(object):
     Example: in "kf5-kio-5.83.0-2.el8.next", the release is "2.el8.next".
     """
 
+    id = attr.ib(type=str, default=None, validator=optional_str, converter=int2str)
```
rohanpm, line 46: May I ask what's the thinking behind storing this as a `str`?  Isn't it always an integer?

tests/koji/test_koji_vmi.py
```diff
@@ -10,6 +10,7 @@
     AmiRelease,
     VMIRelease,
 )
+from pushsource._impl.model.base import KojiBuildInfo
```
rohanpm, line 13: Can you please import it from `pushsource` above like all the rest of the imports?

rohanpm (changes requested):

src/pushsource/_impl/model/base.py
JAVGan: Changed to `int`

tests/koji/test_koji_vmi.py
JAVGan: Done

rohanpm (approved):

PR #356, opened 2023-08-14
Add uefi boot support to compose data [RHELDST-18235]

amcmahon-rh:
From RHEL 8.9 and RHEL 9.3 onwards UEFI will be available. The boot mode must be specified for AMI images to make use of the change. This change adds the boot mode to the metadata parsed in from composes. Data will be used by pubtools-ami and passed to cloudimg

### DIFF
--- a/src/pushsource/_impl/model/ami.py
+++ b/src/pushsource/_impl/model/ami.py
@@ -129,6 +129,11 @@ class AmiPushItem(VMIPushItem):
     )
     """``True`` if the image supports Elastic Network Adapter (ENA)."""
 
+    uefi_support = attr.ib(
+        type=bool, default=None, validator=optional(instance_of(bool))
+    )
+    """``True`` if the image supports UEFI boot."""
+
     billing_codes = attr.ib(
         type=AmiBillingCodes,
         default=None,
@@ -199,6 +204,7 @@ class AmiPushItem(VMIPushItem):
             "description": data["description"],
             "sriov_net_support": data["sriov_net_support"],
             "ena_support": data.get("ena_support"),
+            "uefi_support": data.get("uefi_support"),
             "billing_codes": AmiBillingCodes._from_data(
                 data.get("billing_codes") or {}
             ),
--- a/tests/baseline/cases/staged-simple-ami.yml
+++ b/tests/baseline/cases/staged-simple-ami.yml
@@ -44,6 +44,7 @@ items:
     sriov_net_support: simple
     state: PENDING
     type: access
+    uefi_support: null
     usage_instructions: null
     user_name: null
     version_title: null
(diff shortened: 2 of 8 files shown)


tests/baseline/cases/staged-simple-ami.yml
```diff
@@ -44,6 +44,7 @@ items:
     sriov_net_support: simple
     state: PENDING
     type: access
+    uefi_support: null
```
rohanpm, line 47: Can we please have at least one case where this field isn't using the default/null value?

Without that we're not testing that the value is being read and propagated through the library, e.g. the tests would still pass if you'd forgotten the `"uefi_support": data.get("uefi_support"),` bit or if someone breaks it later.

rohanpm (changes requested):

rbikar (approved):

rohanpm (approved):

src/pushsource/_impl/model/ami.py
```diff
@@ -129,6 +129,11 @@ class AmiPushItem(VMIPushItem):
     )
     """``True`` if the image supports Elastic Network Adapter (ENA)."""
 
+    uefi_support = attr.ib(
+        type=bool, default=None, validator=optional(instance_of(bool))
+    )
+    """``True`` if the image supports UEFI boot."""
+
```
lslebodn, line 136: @amcmahon-rh It is not such simple boolean value.
Image builder provides additional metadata in koji build e.g.
```
 'extra': {'typeinfo': {'image': {'rhel-azure-9.3-20230823.14.aarch64.vhd.xz': {'arch': 'aarch64',
                                                                                'boot_mode': 'uefi'},
                                  'rhel-azure-9.3-20230823.14.x86_64.vhd.xz': {'arch': 'x86_64',
                                                                               'boot_mode': 'hybrid'}}}},
```

aarch64 obviously has to use uefi. But with x86_64 it should try uefi but it can do fallback
in AWS terminology, it is 3 state: "uefi-preferred", "legacy-bios" and "uefi"
* https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ami-boot-mode.html
And one cannot achieve that with boolean in cloudimg
* https://github.com/amcmahon-rh/cloudimg/blob/1ae154580abbeb5ed555d443a2c9bcd61ed1ef70/cloudimg/aws.py#L53

amcmahon-rh: Hey @lslebodn, I don't quite understand, could you elaborate please? 

I understand that AWS requires specific string values. The idea of using a boolean was to keep the value abstract up to the point where we actually interact with AWS, rather than passing string values around different libraries. 

Would this be resolved by having a second boolean? One for uefi and one for bios support?

PR #376, opened 2023-09-14
Change xmlrpc methods to http api calls [RHELDST-17814]

amcmahon-rh:
We want to move away from using xmlrpc methods since they could be used without authentication and are deprecated. This change switches the xmlrpc calls to the equivalent http api call. The change also merges the new errata_http_client into the original client.


rohanpm (changes requested):
While I haven't reviewed all the details, one major issue which sticks out is that the new code *only* supports the HTTP API.

I'm pretty sure that we should be supporting both old and new ways of accessing ET for some time, allowing them to be chosen via an option.  There's a couple of reasons for that.

Firstly for our own sake, imagine that this is deployed and the new client just doesn't work in production.  Things will go much more smoothly if we can leave the code installed but flip a setting to go back to the old client while figuring out the problem.  (or: this even gives us the ability to deploy it without having to wait for ET to fix that weird ftp paths thing.)

Secondly due to the fact that we don't necessarily have control over all the environments where this is used.  I guess you've envisioned that the needed keytab and env vars would be rolled out everywhere before anyone upgrades to this version of the code, but it's not really practical.  This is just a library which anyone can install, use and develop scripts against, and it's not only used within Pub.

Even limiting the scope only to Pub, it's pretty hard to ensure that a change like that will be rolled out across all the weird testing environments, including cases like pipeline-as-a-service.  It wouldn't surprise me at all if some QE environments would prefer to *permanently* keep using the old way, at least as long as the ET code technically allows.

Overall things will go a lot more smoothly if one version of the code supports both old and new ways, and I don't think that costs much code.  I can foresee a future where the old style starts to warn of deprecation and removal by date X, but I don't think the current issue requires going that far.

amcmahon-rh:
@rohanpm  Is this about having xmlrpc and http clients generally available?  The ErrataSource Class makes use of a function that's only available via the http API, so in this case I'd use the HTTP client directly.

I think a factory pattern would be a good approach here. Something like: 

`
USE_HTTP_CLIENT = os.environ.get("PUSHSOURCE_USE_ERRATA_HTTP_API") == "1"

def get_errata_client(threads, url, keytab_path=None, principal=None,
                      force_http=USE_HTTP_CLIENT, **retry_kwargs):
    if force_http:
        return ErrataHTTPClient(threads, url, keytab_path, principal,
                                **retry_kwargs)
    else:
        return ErrataXMLRPCClient(threads, url, **retry_kwargs)
`

rohanpm:
> @rohanpm Is this about having xmlrpc and http clients generally available? The ErrataSource Class makes use of a function that's only available via the http API, so in this case I'd use the HTTP client directly.
> 

Sorry, I didn't really understand what this is asking...

> I think a factory pattern would be a good approach here.

...but sure, something like that code seems just fine.  The main point of discussion would be, what's the default client & what's the logic for deciding between one client and the other.  I'd expect it to default to using the XML-RPC client if the caller didn't provide any keytab.  Having an env var on top of that to force one implementation or the other also sounds OK.

You may be aware already, but if not make sure you see https://github.com/release-engineering/pushsource/pull/379 as well.

If it helps to understand the reason for supporting both.  Imagine this scenario:

- you develop a script/tool using this library to read from ET
- in your requirements you just request 'pushsource', so you're generally using the latest
- you've been using it trouble-free for years
- one day it suddenly starts crashing claiming that some keytab is missing for auth to ET, and nobody has warned you this was coming

A person in that situation would be rightfully pissed off and should question the competence of the pushsource maintainers.

What *should* have happened instead is:

- one day it starts warning like "XML-RPC communication with Errata Tool is deprecated and will be removed in a future version, see (link) for more information"
- user follows link, understands the situation and then updates their config to use non-deprecated communication; might take them some days or weeks if they have to request new service account and so on, during which time they can continue to use the old way.
- Or, user ignores that warning for, let's say 6+ months, maybe *then* one day it starts crashing, this time clearly being the fault of the user's inaction.

That needs one version of pushsource which supports both the old and the new way.

(I don't think the current JIRA issue is actually asking us to add such deprecation warnings, mainly since ET owners themselves never set any date for XML-RPC removal, but it would still be reasonable to add them anyway if you want to set us up for deleting that code a year from now.)

If you're thinking "that scenario can't happen because this is just a library for Pub not used anywhere else, so we have full control over the service where it's used and can ensure things happen in the right order".  I'm explicitly asking for you (all of us) to get out of that mindset and treat this as a generally useful open-source library, for which we as library maintainers are not expected to know or control all the places where it's used.

amcmahon-rh:
I've updated the PR while trying to keep compatibility in mind. A few specific choices and thoughts along the way. 

I've made a base class for the XML-RPC and HTTP clients to inherit from. The XML-RPC version will remain named ErrataClient to avoid any naming and import issues for others using the library. From most users' point of view, it should behave the exact same.

There's now a factory method where we can use an environmental variable to decide which version of the client will be generated. 

This change replicates the new behaviour in https://github.com/release-engineering/pushsource/pull/379 .  Since the XML-RPC version of the class doesn't override the `get_advisory_data` function, it will always return None. If you're using the XML-RPC client, this data wouldn't be available anyway and I feel like this approach makes the clients more interchangeable. 

The change in https://github.com/release-engineering/pushsource/pull/379 did make sense, since the original HTTP client had just one function returning "optional" data.  In this case though, if you specify the HTTP client and don't pass authentication, all the functions will fail, you'd get no data, and everything will eventually break. Better to let it fail with authentication errors.  


LMK if anything I've said doesn't make sense and I'll try to explain it better.

src/pushsource/_impl/backend/errata_source/errata_client.py
```diff
@@ -1,16 +1,45 @@
...
 from more_executors.futures import f_zip, f_map
 
 from ...compat_attr import attr
 
+
+
 LOG = logging.getLogger("pushsource.errata_client")
 
 
+USE_HTTP_CLIENT = os.environ.get("PUSHSOURCE_USE_ERRATA_HTTP_API") is not None
```
rohanpm, line 25: I don't think it should be "is not None" because then even setting the variable to a blank string or to "0" will cause it to be enabled, which is unusual.

It should probably be `== "1"`.

src/pushsource/_impl/backend/errata_source/errata_client.py
```diff
@@ -73,49 +107,197 @@ def get_raw_f(self, advisory_id):
...
             raise
+
+
+class ErrataHTTPClient(ErrataClientBase):
+    def __init__(self, threads, url, keytab_path: str = None,
+                 principal: str = None, **retry_args):
+        super().__init__(threads, url, **retry_args)
+
+        self.get_advisory_data = partial(
+            self._call_et, "/api/v1/push_metadata/ftp_paths/{id}.json"
```
rohanpm, line 164: I don't get this, why is get_advisory_data using ftp_paths?  Doesn't it need to use /api/v1/erratum/{id} like it did before this change?

Also, if this is a mistake, were tests somehow unable to catch it?  If so they should be improved.

src/pushsource/_impl/backend/errata_source/errata_client.py
```diff
@@ -73,49 +107,197 @@ def get_raw_f(self, advisory_id):
...
+                                              None)
+
+        if principal:
+            self.principal = principal
+        else:
+            self.principal = os.environ.get("PUSHSOURCE_ERRATA_PRINCIPAL",
+                                            None)
+
+    def authenticate(self):
+        # Shunting like this for compatibility, since this class has probably
```
rohanpm, line 196: While I don't have any problem with the design here, I'm not quite following what this comment means.
In case there is some misunderstanding, I want to point out that nothing outside of pushsource is expected to have access to the ErrataHTTPClient class.  It's not a part of the library's API.

rohanpm (changes requested):
Overall looks pretty good, it needs a rebase to fix conflicts though.

src/pushsource/_impl/backend/errata_source/errata_client.py
amcmahon-rh: Is this the same for ErrataClient? I just wanted to make these classes interchangeable. I assumed ErrataHTTPClient could be accessed outside of PushSource and that it may have been used already. If that's not the case I can just put the kerberos ticket code under authentication

src/pushsource/_impl/backend/errata_source/errata_client.py
amcmahon-rh: Bad copy paste job. I wanted to get opinions on this approach before going through and making all the tests work.

src/pushsource/_impl/backend/errata_source/errata_client.py
rohanpm: Yes, it's the same.

If you look at https://github.com/release-engineering/pushsource/blob/master/src/pushsource/__init__.py , only those classes are the ones which can be accessed via `import pushsource`.

For everything else, there is no way to access it without writing `from pushsource._impl.(something)`, and if developers write that code they're supposed to understand from the naming that they're accessing something private & shouldn't be doing that

src/pushsource/_impl/backend/errata_source/errata_client.py
```diff
@@ -1,16 +1,43 @@
...
+
+def get_errata_client(
+    threads,
+    url,
+    keytab_path=None,
+    principal=None,
+    force_http=USE_HTTP_CLIENT,
+    **retry_kwargs,
+):
+    if force_http:
```
rohanpm, line 35: Should it be also trying to use the HTTP client if keytab/principal were provided?

If it doesn't, I think that would already break the `product_name` support added recently.  Because the implemented behavior was, "you get product_name if you pass keytab/principal", but with this logic as you've got it here it'd instead be "you get product_name only if PUSHSOURCE_USE_ERRATA_HTTP_API is set to 1".

amcmahon-rh: I had intended for the `PUSHSOURCE_USE_ERRATA_HTTP_API` to replace product_name values. Given what you mentioned before about the variability of environments, I liked the idea of having some sort of "handbrake". Maybe it can be flipped around, so it'll use HTTP by default if keytab/principal are provided, but we have an envvar that can force it to use xmlrpc

rohanpm: I don't understand what "replace product_name values" means there.

Let me put it this way.  This is what I expect the state of the world to be leading up to deployment of this change:

1. Some environments using this library are providing keytab/principal, and are expecting to see non-null `product_name`.
2. Some environments using this library are not providing keytab/principal, and don't care about `product_name`.

We must not break either of those cases.

i.e. when upgrading to the version of pushsource with this change, in case (1), `product_name` must not suddenly disappear, and in case (2), keytab/principal must not suddenly become mandatory.

The most obvious way to achieve this seems to be making it default to the HTTP client (which supports `product_name`) if keytab/principal is provided, and default to the XML-RPC client if keytab/principal is not provided.

An env var on top of this to force one or the other is fine but I'd see that as being only for emergency usage, if something isn't working how we expect it to after deployment.

rohanpm (commented):
I don't see any obvious reason not to proceed with this, but the PR description still says it's a WIP.  Please let us know when the change is ready for final review.

amcmahon-rh:
@rohanpm  and @rbikar  ptal

src/pushsource/_impl/backend/errata_source/errata_client.py
```diff
@@ -73,49 +110,193 @@ def get_raw_f(self, advisory_id):
...
                         "advisory": advisory_id,
                     }
                 },
             )
             raise
+
+
+class ErrataClient(ErrataClientBase):
+    def __init__(self, threads, url, **retry_args):
+        deprecation_notice = """
```
rohanpm, line 163: I'm pretty sure all the whitespace in this string is going to be logged verbatim, which will make it look very messy and wastes a lot of horizontal space.

You probably want to do something like:

```python
deprecation_notice = textwrap.dedent("""
     The XML RPC Errata Client has been deprecated and will be removed.
     Some more info here, and so on.
""").strip()
```

Or even just avoid any line breaks in the final string, like:

```python
deprecation_notice = (
  "The XML RPC Errata Client has been deprecated and will be removed. "
  "Some more info here, "
  "and even some more here, none of which will split over multiple lines."
)
```

src/pushsource/_impl/backend/errata_source/errata_client.py
```diff
@@ -73,49 +110,193 @@ def get_raw_f(self, advisory_id):
...
+               The XML RPC Errata Client has been deprecated and will be removed.
+               Please see <some link> for more information on how to transition to
```
rohanpm, line 165: The "some link" was meant to be a placeholder which would be pointed at some actual doc we'd write about this.

That being said though, maybe the situation is not really complicated enough to warrant that.  I guess the instructions are simple enough to go in a single sentence? Here's my attempt:

```
The XML RPC Errata Client has been deprecated and will be removed.
Please provide a `keytab_path` and `principal` to the errata pushsource backend.
```

src/pushsource/_impl/backend/errata_source/errata_client.py
```diff
@@ -73,49 +110,193 @@ def get_raw_f(self, advisory_id):
...
+               the HTTP client.  
+               """
+        logging.warning(deprecation_notice)
```
rohanpm, line 168: While I'm not super familiar with it myself, I'm wondering if we should be using `warnings` here rather than `logging`.

```
warnings.warn(deprecation_notice, DeprecationWarning)
```

The reason I say this is because `warnings` allows the user to suppress this warning if they already know about it, or to configure it so that it only appears once per run - https://docs.python.org/3/library/warnings.html#the-warnings-filter .  

Without this, a user who is - for example - running this code 1000 times, will see this warning 1000 times, possibly overwhelming their usual logs, and they'll have no reasonable way to suppress it.

rohanpm (changes requested):

docs/sources/staged.rst
```diff
@@ -2,7 +2,7 @@ Source: staged
 ==============
 
 The ``staged`` push source allows the loading of content from any locally accessible
-directory referred to as the staging area. Content types and destinations are defined
+directory referred to as the staging area. Content types and destinations s and destinations are defined
```
rohanpm, line 5: Looks like you added some junk into your commit, can you please clean it up?

I'm not sure what your workflow is, but if you are currently doing something like `git commit -a` you might want to give `git add -p` a try which allows you to review every chunk you're adding into your commits.

The change looks good apart from this.

rohanpm (changes requested):

docs/sources/staged.rst
amcmahon-rh: I have a "quick amend" git alias, which is usually fine, but I think I must have fat fingered that docs file while researching  kickstarts.  The `-p` arg sounds handy though :+1:

rohanpm (approved):

rbikar (approved):

PR #379, opened 2023-09-19
Allow Errata HTTP client to work without Kerberos credentials

querti:
If credentials are not specified, the HTTP method returns None without querying the endpoint. This change was made because the Errata Kerberos credentials are not yet prepared and the gathered information is not yet necessary.

This unusual feature should be removed in the future.

### DIFF
--- a/src/pushsource/_impl/backend/errata_source/errata_http_client.py
+++ b/src/pushsource/_impl/backend/errata_source/errata_http_client.py
@@ -5,6 +5,7 @@ import re
 import logging
 import tempfile
 from urllib.parse import urljoin
+from functools import wraps
 
 import requests
 import gssapi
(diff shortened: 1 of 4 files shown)


querti:
Without this fix, container pushes will fail...

midnightercz:
I don't understand when exactly it can happen that credentials are not ready yet. Can you be more specific?

querti:
> I don't understand when exactly it can happen that credentials are not ready yet. Can you be more specific?

The ET team haven't provided us with a keytab.

rohanpm:
> The ET team haven't provided us with a keytab

That comment doesn't make sense to me, it would not be ET team's responsibility to provide us with a keytab.

The actual change makes sense to me though simply from a compatibility perspective.  Since this library didn't originally need any keytab here, 100% of environments using the library would initially be missing that; and since this is just a library that anyone can install and use without asking our permission, we also don't have control over every environment where it's used, so we cannot ensure that everyone will provide a keytab in a specific timeframe.

We should try to avoid creating "traps" where simply upgrading a library causes something to crash because it suddenly needs something it's never needed before.  Looks like we created such a trap before, and this commit is removing it which is good.

rohanpm (approved):

PR #463, opened 2024-02-20
Adds to AMI Access Endpoint URL used for aws marketplace

jajreidy:
Adds to AMI ability to add access endpoint url. Used for version creation for some products in AWS Marketplace.

Refers to SPSTRAT-126

### DIFF
--- a/src/pushsource/_impl/model/ami.py
+++ b/src/pushsource/_impl/model/ami.py
@@ -93,6 +93,27 @@ class AmiSecurityGroup(object):
 
 
 @attr.s()
+class AmiAccessEndpointUrl(object):
+    """Access Endpoint URL to be associated with a Marketplace VM."""
+
+    port = attr.ib(type=int, validator=instance_of(int))
+    """Port to access the endpoint URL."""
+
+    protocol = attr.ib(type=str, validator=instance_of_str)
+    """Protocol to access the endpoint URL (http, https)."""
+
+    @classmethod
+    def _from_data(cls, data):
+        """Instantiate SecurityGroup from raw dict"""
+
+        kwargs = {
+            "port": data["port"],
+            "protocol": data["protocol"],
+        }
+        return cls(**kwargs)
+
+
+@attr.s()
 class AmiPushItem(VMIPushItem):
     """A :class:`~pushsource.PushItem` representing an Amazon Machine Image (or "AMI").
 
@@ -180,6 +201,13 @@ class AmiPushItem(VMIPushItem):
     )
     """Automatically created security groups for the product. """
 
+    access_endpoint_url = attr.ib(
+        type=AmiAccessEndpointUrl,
+        default=None,
+        validator=optional(instance_of(AmiAccessEndpointUrl)),
+    )
+    """Billing codes associated with this image."""
+
     @classmethod
     def _from_data(cls, data):
         """Instantiate AmiPushItem from raw list or dict"""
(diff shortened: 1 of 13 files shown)


jajreidy:
@lslebodn @JAVGan PTAL as well

src/pushsource/_impl/schema/staged-schema.yaml
```diff
@@ -173,6 +173,18 @@ definitions:
...
+                type:
+                - object
+                - "null"
+                properties:
+                    port:
+                        type: number
+                        minimum: 0
+                        maximum: 65536
+                    protocol:
+                        enum: ["http", "https"]
```
JAVGan, line 186: Did you confirm the supported types are just these in some docs? Or are we just following the logic here as the endpoint must be a web one?

jajreidy: I checked via the WebUI when creating a version.

JAVGan:
Just adding a bit of context: this will be required for some AMIs with certain products that has administrative Web pages exposed as an endpoint, like JBoss EAP.

JAVGan (approved):

src/pushsource/_impl/model/ami.py
```diff
@@ -92,6 +92,27 @@ def _from_data(cls, data):
...
 
 
+@attr.s()
+class AmiAccessEndpointUrl(object):
+    """Access Endpoint URL to be associated with a Marketplace VM."""
+
+    port = attr.ib(type=int, validator=instance_of(int))
+    """Port to access the endpoint URL."""
+
+    protocol = attr.ib(type=str, validator=instance_of_str)
```
rohanpm, line 102: Minor: I'm not sure if you are copying some existing AWS API which already uses the term "protocol".  If so, ignore this.

But in the context of a URL, this part of a URL is referred to as a "scheme", not "protocol".  For example the named tuple returned by https://docs.python.org/3/library/urllib.parse.html uses "scheme".  Given that this class also represents a URL, it would be nice to be consistent with that.

src/pushsource/_impl/model/ami.py
```diff
@@ -92,6 +92,27 @@ def _from_data(cls, data):
...
+    """Protocol to access the endpoint URL (http, https)."""
+
+    @classmethod
+    def _from_data(cls, data):
+        """Instantiate SecurityGroup from raw dict"""
```
rohanpm, line 107: False copy-pasted comment, this isn't a SecurityGroup.

Deleting a comment entirely is preferred over copy-pasting an incorrect one.

src/pushsource/_impl/model/ami.py
```diff
@@ -180,6 +201,13 @@ class AmiPushItem(VMIPushItem):
     )
     """Automatically created security groups for the product. """
 
+    access_endpoint_url = attr.ib(
+        type=AmiAccessEndpointUrl,
+        default=None,
+        validator=optional(instance_of(AmiAccessEndpointUrl)),
+    )
+    """Billing codes associated with this image."""
```
rohanpm, line 209: Another false copy-pasted comment/doc string (this will be part of the public docs so it has to be fixed).

src/pushsource/_impl/schema/staged-schema.yaml
rohanpm: Related to my other comment about "scheme" vs "protocol", does this follow some existing API which uses "protocol"?

src/pushsource/_impl/model/ami.py
jajreidy: same here

rohanpm (approved):

PR #464, opened 2024-02-21
Add attribute "marketplace_title" to VMIRelease

JAVGan:
This commit introduces a new optional parameter on `VMIRelease` and `AmiRelease` models called `marketplace_title` which is going to be used by the pubtool as the marketplace "version title" whenever it's set, by overriding the usual version title which the pubtool was going to compute.

This is required for some layered products which requires a specialized templating title instead of the one generated by the pubtool.

With this, it will be possible to transport the template via the VMIPushItem from the source to the pubtooling which will format it into the apropriate value for the marketaplace.

Refers to SPSTRAT-116

### DIFF
--- a/src/pushsource/_impl/model/vms.py
+++ b/src/pushsource/_impl/model/vms.py
@@ -3,6 +3,7 @@ import enum
 
 from .base import PushItem
 from .. import compat_attr as attr
+from attr import asdict
 from .conv import datestr, instance_of_str, instance_of, optional_str, optional, in_
 
 
@@ -73,6 +74,22 @@ class VMIPushItem(PushItem):
 
     boot_mode = attr.ib(type=BootMode, default=None, validator=optional(in_(BootMode)))
     """Boot mode supported by the image (if known): uefi, legacy, or hybrid (uefi + legacy)."""
+
+    marketplace_title_template = attr.ib(type=str, default=None, validator=optional_str)
+    """The template is of the form used by ``str.format``, with available keywords being all of
+    the documented fields on ``VMIRelease`` and ``AMIRelease`` classes.
+    
+    It's used by the property `marketplace_title` to format it as the marketplace title."""
+
+    @property
+    def marketplace_title(self) -> str:
+        """The marketplace title which is used for some certain layared products.
+
+        It's built from the `marketplace_title_template` by formatting it with the proper values.
+        """
+        if not self.marketplace_title_template or not self.release:
+            return ""
+        return self.marketplace_title_template.format(**asdict(self.release))
 
     @release.validator
     def __validate_release(self, attribute, value):  # pylint: disable=unused-argument
--- a/src/pushsource/_impl/model/vms.py
+++ b/src/pushsource/_impl/model/vms.py
@@ -83,7 +83,7 @@ class VMIPushItem(PushItem):
 
     @property
     def marketplace_title(self) -> str:
-        """The marketplace title which is used for some certain layared products.
+        """The marketplace title which is used for some certain layered products.
 
         It's built from the `marketplace_title_template` by formatting it with the proper values.
         """
--- a/tests/baseline/cases/staged-simple-ami-bc.yml
+++ b/tests/baseline/cases/staged-simple-ami-bc.yml
@@ -23,6 +23,7 @@ items:
     ena_support: true
     image_id: null
     marketplace_entity_type: AMIProduct
+    marketplace_title_template: null
     md5sum: null
     name: fake-image.raw
     origin: {{ src_dir }}/tests/staged/data/simple_ami_with_bc
--- a/src/pushsource/_impl/model/ami.py
+++ b/src/pushsource/_impl/model/ami.py
@@ -220,6 +220,8 @@ class AmiPushItem(VMIPushItem):
             "scanning_port": data.get("scanning_port") or None,
             "user_name": data.get("user_name") or None,
             "version_title": data.get("version_title") or None,
+            "marketplace_title_template": data.get("marketplace_title_template")
+            or None,
             "security_groups": [
                 AmiSecurityGroup._from_data(security_group)
                 for security_group in (data.get("security_groups") or [])
--- a/src/pushsource/_impl/schema/staged-schema.yaml
+++ b/src/pushsource/_impl/schema/staged-schema.yaml
@@ -135,6 +135,8 @@ definitions:
                 $ref: "#/definitions/optional_string"
             version_title:
                 $ref: "#/definitions/optional_string"
+            marketplace_title_template:
+                $ref: "#/definitions/optional_string"
             security_groups:
                 type:
                     - array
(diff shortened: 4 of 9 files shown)


JAVGan:
@lslebodn @jajreidy @rohanpm PTAL

jajreidy:
LGTM

tests/baseline/cases/staged-simple-ami-bc.yml
```diff
@@ -34,6 +34,7 @@ items:
       base_product: null
       base_version: null
       date: 2020-05-11
+      marketplace_title: null
```
rohanpm, line 37: Could you please make at least one of the tests have a non-null value?

Otherwise the tests would pass even if the loading code was broken and always set the value to None.

rohanpm (changes requested):

src/pushsource/_impl/model/vms.py
```diff
@@ -54,6 +54,10 @@ class VMIRelease(object):
     type = attr.ib(type=str, default=None, validator=optional_str)
     """Release type, typically "ga" or "beta"."""
 
+    marketplace_title = attr.ib(type=str, default=None, validator=optional_str)
+    """Customer facing title or title template for the marketplace release. For example,
```
rohanpm, line 58: Is there something which can be said more specific than "template"?  Like, "jinja-compatible template" if that's what it is?

rohanpm: ... or: template intended for usage with `str.format`.  But it kind of seems like the variables need to be defined somewhere.  Isn't it going to crash if you try to render the format without providing all of the variables it uses?

The JIRA issue also doesn't explain anything about it.  I don't like the idea of exposing some kind of template of an undefined format, a year or two down the line we could be arguing about which component is buggy if the template produced by one tool doesn't meet the format expected by some other tool.

JAVGan: Yet another way is to use some regex to delimit our "templates", like:

"This will be the version @VERSION@ of product @PRODUCT@" and we do the formatting using some kind of regex/replace.

JAVGan: But supposing we agree that the accepted variables may be any attribute of `VMIRelease` / `AmiRelease`.

The following snippet could be a way to format the string by any tooling:

```
string_to_format = push_item.marketplace_title
keywords = re.findall(r'{\w+}', string_to_format)
params = {}
for k in keywords:
    kw = k.replace("{", "").replace("}", "")
    params[kw] = getattr(pushitem.release, kw, "")

return string_to_format.format(**params)
```

WDYT?

tests/baseline/cases/staged-simple-ami-bc.yml
JAVGan: Test added.

src/pushsource/_impl/model/vms.py
rohanpm: According to a quick test I did just now, str.format will happily ignore any unused keywords without generating a warning or error.  So could the above code be simplified more to this? (Assuming the caller is willing to call a method from `attr` library)

```
push_item.marketplace_title.format(**attr.asdict(pushitem.release))
```

But going that far, I start to wonder if this attribute should rather be named `marketplace_title_template` and then `marketplace_title` is a `@property` which more or less contains the above code?

Anyway, I like the following two options, documentation-wise:

> The template is of the form used by `str.format`, with available keywords being all of the documented fields on `VMIRelease` and `AMIRelease` classes.

Or:

> The template is of the form used by (some reference/link to some other project)

For the latter approach I wouldn't care much about the details, just that "something" somewhere is responsible for defining the format, so that we know it's not the responsibility of pushsource library.

JAVGan: @rohanpm I've changed it completely to follow your suggestion:

- Use the `marketplace_title_template` in `VMIPushItem` to receive any variables from `VMIRelease`
- Create the property `marketplace_title` in `VMIRelease` to format the template
- Updated tests

Please let me know if it's good now

src/pushsource/_impl/model/vms.py
```diff
@@ -74,6 +74,16 @@ class VMIPushItem(PushItem):
     boot_mode = attr.ib(type=BootMode, default=None, validator=optional(in_(BootMode)))
     """Boot mode supported by the image (if known): uefi, legacy, or hybrid (uefi + legacy)."""
 
+    marketplace_title_template = attr.ib(type=str, default=None, validator=optional_str)
+    """The template is of the form used by ``str.format``, with available keywords being all of
```
rohanpm, line 78: The doc string should still say something about what the attribute actually means, not just what the format is.

Given that there is a `marketplace_title` property, this attribute's doc string could mention that it's used to generate that value, and a sentence explaining what `marketplace_title` actually means could be in that property's doc string.

src/pushsource/_impl/model/vms.py
```diff
@@ -74,6 +74,16 @@ class VMIPushItem(PushItem):
...
+    the documented fields on ``VMIRelease`` and ``AMIRelease`` classes."""
+
+    @property
+    def marketplace_title(self) -> str:
+        if not self.marketplace_title_template:
```
rohanpm, line 83: It seems that `self.release` is allowed to be `None`, so should that also cause it to go into this branch and return an empty string?

src/pushsource/_impl/model/vms.py
```diff
@@ -74,6 +74,16 @@ class VMIPushItem(PushItem):
...
+            return ""
+        return self.marketplace_title_template.format(**attr.attr.asdict(self.release))
```
rohanpm, line 85: Minor - the double reference to `attr.attr` is a little weird and inconsistent with what currently exists.

The idea has been that anything used within `attr` is imported to `compat_attr` and then exported again.  That means if we drop support for old python/attr versions, at some point we can just replace `from .. import compat_attr as attr` with `import attr` everywhere and not have to touch anything else.  But it won't work here as the pattern was not followed.

rohanpm (changes requested):

src/pushsource/_impl/model/vms.py
JAVGan: I've changed it, PTAL again

src/pushsource/_impl/model/vms.py
JAVGan: You're right, good catch! I've patched it accordingly.

src/pushsource/_impl/model/vms.py
JAVGan: I've changed it, PTAL again

src/pushsource/_impl/model/vms.py
```diff
@@ -74,6 +75,22 @@ class VMIPushItem(PushItem):
...
 
+    marketplace_title_template = attr.ib(type=str, default=None, validator=optional_str)
+    """The template is of the form used by ``str.format``, with available keywords being all of
+    the documented fields on ``VMIRelease`` and ``AMIRelease`` classes.
+    
+    It's used by the property `marketplace_title` to format it as the marketplace title."""
+
+    @property
+    def marketplace_title(self) -> str:
+        """The marketplace title which is used for some certain layared products.
```
rohanpm, line 86:
```suggestion
        """The marketplace title which is used for some certain layered products.
```

rohanpm (approved):

PR #467, opened 2024-02-29
adding support for art build

ashwini3326:
This merge provides support for generating push items from ART/rhcos builds.

### DIFF
--- a/src/pushsource/_impl/backend/koji_source.py
+++ b/src/pushsource/_impl/backend/koji_source.py
@@ -520,7 +520,7 @@ class KojiSource(Source):
             message = "Virtual machine image build not found in koji: %s" % nvr
             LOG.debug(message)
             raise ValueError(message)
-        
+
         # VMI has the {'typeinfo': {'image':{}}} on extra
         extra = meta.get("extra") or {}
         typeinfo = extra.get("typeinfo") or {}
@@ -537,7 +537,7 @@ class KojiSource(Source):
             message = "Build %s not recognized as a virtual machine image build" % nvr
             LOG.debug(message)
             raise ValueError(message)
-        
+
         build_id = meta["id"]
         archives = self._get_archives(build_id)
         for archive in archives:
@@ -554,7 +554,7 @@ class KojiSource(Source):
 
         for k in azure_types:
             cloud_types.setdefault(k, VHDPushItem)
-        
+
         # Collect all known VM image archives
         vmi_archives = [
             elem
@@ -621,7 +621,7 @@ class KojiSource(Source):
             )
 
         return out
-    
+
     def _push_items_from_rhcos_build(self, nvr, meta, archives=None):
 
         LOG.debug("Looking for core os assembler build on %s, %s", nvr, meta)
@@ -639,17 +639,17 @@ class KojiSource(Source):
         extra = meta.get("extra") or {}
         image = extra.get("typeinfo").get("image") or {}
 
-        arch = image.get("arch") 
+        arch = image.get("arch")
 
         boot_mode = image.get("boot_mode")
         if boot_mode is not None:
             boot_mode = BootMode(boot_mode)
 
-        rhcos_container_config = rhcos_meta_data.get("coreos-assembler.container-config-git") or {}
+        rhcos_container_config = (
+            rhcos_meta_data.get("coreos-assembler.container-config-git") or {}
+        )
         branch = rhcos_container_config.get("branch") or ""
         version = branch.split("-")[-1] or ""
-
-
 
         # Try to get the resping version from release, if available
         respin = try_int(meta["release"])
@@ -665,9 +665,7 @@ class KojiSource(Source):
                     "custom_meta_data": {
                         "region": ami.get("name"),
                         "src": ami.get("hvm"),
-
-                    }
-
+                    },
                 }
             )
 
@@ -678,9 +676,7 @@ class KojiSource(Source):
                 "push_item_class": VHDPushItem,
                 "custom_meta_data": {
                     "src": rhcos_meta_data["azure"]["url"],
-
-                }
-
+                },
             }
         )
 
@@ -712,7 +708,7 @@ class KojiSource(Source):
                         version=version,
                         release=meta["release"],
                     ),
-                    sha256sum = sha256sum,
+                    sha256sum=sha256sum,
                     release=release,
                     marketplace_name=vm_item["marketplace_name"]
                 )
--- a/src/pushsource/_impl/backend/koji_source.py
+++ b/src/pushsource/_impl/backend/koji_source.py
@@ -2,6 +2,7 @@ import os
 import threading
 import logging
 from functools import partial
+import json
 
 from queue import Queue, Empty
 from threading import Thread
@@ -519,7 +520,7 @@ class KojiSource(Source):
             message = "Virtual machine image build not found in koji: %s" % nvr
             LOG.debug(message)
             raise ValueError(message)
-
+        
         # VMI has the {'typeinfo': {'image':{}}} on extra
         extra = meta.get("extra") or {}
         typeinfo = extra.get("typeinfo") or {}
@@ -536,10 +537,12 @@ class KojiSource(Source):
             message = "Build %s not recognized as a virtual machine image build" % nvr
             LOG.debug(message)
             raise ValueError(message)
-
+        
         build_id = meta["id"]
-
         archives = self._get_archives(build_id)
+        for archive in archives:
+            if archive["filename"] == "meta.json":
+                return self._push_items_from_rhcos_build(nvr, meta)
 
         # Supported types of VMI archives
         ami_types = ["raw", "raw-xz"]
@@ -551,7 +554,7 @@ class KojiSource(Source):
 
         for k in azure_types:
             cloud_types.setdefault(k, VHDPushItem)
-
+        
         # Collect all known VM image archives
         vmi_archives = [
             elem
@@ -559,9 +562,7 @@ class KojiSource(Source):
             if elem.get("btype") == "image"
             and elem.get("type_name") in cloud_types.keys()
         ]
-
         out = []
-
         # Generate the respective PushItem type for each archive
         for archive in vmi_archives:
             path = self._pathinfo.typedir(meta, archive["btype"])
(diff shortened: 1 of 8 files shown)


src/pushsource/_impl/backend/koji_source.py
```diff
@@ -621,6 +622,92 @@ def _push_items_from_vmi_build(self, nvr, meta):
...
+                    **vm_item["custom_meta_data"],
+                    boot_mode=boot_mode,
+                    build=nvr,
+                    build_info=KojiBuildInfo(
+                        id=int(build_id),
+                        name=meta["name"],
+                        version=meta["version"],
+                        release=meta["release"],
+                    ),
+                    sha256sum=rhcos_meta_data.get("images")
```
rohanpm, line 702: This will crash badly if any of the sums are missing in this "image" structure.  Is it acceptable?

tests/koji/test_koji_vmi.py
```diff
@@ -238,3 +239,140 @@ def test_koji_vmi_compound_product_name(fake_koji, koji_dir):
...
+        "url": "https://example.windows.net/imagebucket/rhcos-414.92-0-azure.x86_64.vhd"},
+ "amis": [
+    {"name": "us-gov-west-1",
+   "hvm": "ami-01"
+   },
+    {"name": "us-east-1",
+   "hvm": "ami-02"
+   }]}"""
+
+    patch.object(os.path, "join", return_value=path)
```
rohanpm, line 321: This mocking doesn't seem like it should be necessary, because you can easily point the koji source to point at any directory you want.  In fact you're already pointing it at `koji_dir`, so you just need to write the meta.json file to the correct path under that dir and it should work without mocking.

Can you please do it that way?  This way is really nasty, because:

1) You're not checking that 'join' is used correctly at all.  The test can pass even if the code is totally broken and calls 'join' with incorrect arguments.

2) Mocking like this requires more maintenance in the log term, because perfectly correct changes in the code can cause the test to break just by not calling exactly the methods that you've mocked here.

rohanpm (changes requested):
This looks good, but some coverage is missing.  Could you please find a way to get it to 100% coverage?

src/pushsource/_impl/backend/koji_source.py
rohanpm: If you have easy access to it then I'd suggest including the sha256sum even without a specific use-case in your own tools.  It's one of the attributes on the PushItem base class and Pub understands how to index it, so if you produce this attribute (and do a few other things right) then you can search for push tasks involving that checksum, which can be useful when handling tickets.

ashwini3326: PTAL

tests/koji/test_koji_vmi.py
ashwini3326: I have removed it, PTAL

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -538,8 +539,10 @@ def _push_items_from_vmi_build(self, nvr, meta):
             raise ValueError(message)
 
         build_id = meta["id"]
-
         archives = self._get_archives(build_id)
+        for archive in archives:
+            if archive["filename"] == "meta.json":
+                return self._push_items_from_rhcos_build(nvr, meta)
```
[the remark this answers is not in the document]
ashwini3326: implemented

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -621,6 +622,97 @@ def _push_items_from_vmi_build(self, nvr, meta):
...
+                with open(item_src) as archive_file:
+                    rhcos_meta_data = json.loads(archive_file.read())
+
+        extra = meta.get("extra") or {}
+        image = extra.get("typeinfo").get("image") or {}
+
+        arch = image.get("arch")
+        boot_mode = image.get("boot_mode")
+
+        version = rhcos_meta_data["coreos-assembler.container-config-git"][
```
rohanpm, line 645: Is this structure documented anywhere?

This code doesn't tolerate a missing field at all, it'll just crash if the field isn't there.  If the owners of the tool generating this field haven't published a doc with the fields they're producing, I'm concerned that they'll feel free to change it at any time.

The concern would be mitigated if this could be written in such a way that it wouldn't crash.

tests/koji/test_koji_vmi.py
```diff
@@ -238,3 +239,144 @@ def test_koji_vmi_compound_product_name(fake_koji, koji_dir):
...
+        "url": "https://example.windows.net/imagebucket/rhcos-4.11-0-azure.x86_64.vhd"},
+ "amis": [
+    {"name": "us-gov-west-1",
+   "hvm": "ami-01"
+   },
+    {"name": "us-east-1",
+   "hvm": "ami-02"
+   }]}"""
+
+    with patch("builtins.open", mock_open(read_data=meta_data)) as mocked_open:
```
rohanpm, line 326: What I wrote in comment https://github.com/release-engineering/pushsource/pull/467#discussion_r1508205061 applies here too.  Please, don't mock `open`, it's completely unnecessary and it makes the test fragile as it now makes the test depend on the code-under-test calling `open` exactly once and only for this purpose.  If someone needs to add code later which opens something else, they will be forced to rewrite this test.

There should be no reason to mock any `open` or `path` methods here.  Please, check the value of the `koji_dir` variable while this test runs.  It ought to be a directory that you can write to.  Just arrange for the JSON file to really exist under that directory and then let the code under test really read it from that directory.  Adding mocks unnecessarily makes the tests more complicated and fragile.

rohanpm (changes requested):

src/pushsource/_impl/backend/koji_source.py
ashwini3326: https://github.com/coreos/coreos-assembler/blob/efa00a6b07431bbfdf7a80032612290aea6a62bc/src/v1.json#L244
This file have the schema for rhcos meta.json file, but it seems that they have marked "coreos-assembler.container-config-git" as optional. I have given one 'or' block. let me know if it looks good or not.

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -621,6 +622,97 @@ def _push_items_from_vmi_build(self, nvr, meta):
...
+                    description=rhcos_meta_data.get("summary"),
+                    **vm_item["custom_meta_data"],
+                    boot_mode=boot_mode,
+                    build=nvr,
+                    build_info=KojiBuildInfo(
+                        id=int(build_id),
+                        name=meta["name"],
+                        version=version,
+                        release=meta["release"],
+                    ),
```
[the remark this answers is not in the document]
ashwini3326: I changed it to use the name of marketplace it supposed to be shipped. I did it for testing it properly. But if  we agree on other approach, I can implement it.

JAVGan (approved):

src/pushsource/_impl/backend/koji_source.py
```diff
@@ -621,6 +622,100 @@ def _push_items_from_vmi_build(self, nvr, meta):
...
+        image = extra.get("typeinfo").get("image") or {}
+
+        arch = image.get("arch")
+
+        boot_mode = image.get("boot_mode")
+        if boot_mode is not None:
+            boot_mode = BootMode(boot_mode)
+
+        version = (
+            rhcos_meta_data["coreos-assembler.container-config-git"]["branch"].split(
```
rohanpm, line 649: You mentioned changing this to cope with the fact that this field is optional in the schema, but this is still going to crash if the field is absent or `None`.  Adding some test data like that could confirm it if needed.

If you want to make this tolerant against the missing field, won't it have to be written something more like this?

```python
branch = (rhcos_meta_data.get("coreos-assembler.container-config-git") or {}).get("branch") or ""
version = branch.split("-")[-1]
```

tests/koji/test_koji_vmi.py
```diff
@@ -1,6 +1,7 @@
 import os
 
 from pytest import raises, mark
+from unittest.mock import patch, mock_open
```
rohanpm, line 4: These imports aren't used, can you please clean them up?

rohanpm (changes requested):

src/pushsource/_impl/backend/koji_source.py
ashwini3326: my bad I missed it, sending it again.

ashwini3326:
I need to change the naming convention for  rhcos images. I am working for it, will publish my changes once it gets completed.

tests/koji/test_koji_vmi.py
ashwini3326: removed

rohanpm (dismissed):

JAVGan (approved):

rohanpm:
> I need to change the naming convention for rhcos images

@ashwini3326 could you please confirm if this PR is complete and OK to release, or are there still changes expected?

ashwini3326:
@rohanpm  After some discussion on the way we are currently naming the rhcos packages, I need to change few things and also add a 'marketplace_target' metadata for VMs'.  so we need to wait before it gets merged.

ashwini3326:
@rohanpm as Red Hat Core OS images are named differently ( it only have the name rhcos-<arch>) we need a way to identify where these images are expected to be shipped. I am adding a new attribute (marketplace_name) to VMIPushItem for this purpose. PTAL. 

Also let me know if we need more tests for this change.

JAVGan:
@ashwini3326 please run the tests again with `UPDATE_BASELINES=1` and include them into this PR as the tests are failing due to missing the new attribute `marketplace_name`

```
E       AssertionError: Output differs from baseline:
E       
E       --- 
E       +++ 
E       @@ -24,6 +24,7 @@
E            ena_support: true
E            image_id: null
E            marketplace_entity_type: AMIProduct
E       +    marketplace_name: null
E            marketplace_title_template: null
E            md5sum: null
E            name: fake-image.raw
E       
E       
E       Re-run with PUSHSOURCE_UPDATE_BASELINE=1 if this is correct.
```

ashwini3326:
> @ashwini3326 please run the tests again with `PUSHSOURCE_UPDATE_BASELINES=1` and include them into this PR as the tests are failing due to missing the new attribute `marketplace_name`
> 
> ```
> E       AssertionError: Output differs from baseline:
> E       
> E       --- 
> E       +++ 
> E       @@ -24,6 +24,7 @@
> E            ena_support: true
> E            image_id: null
> E            marketplace_entity_type: AMIProduct
> E       +    marketplace_name: null
> E            marketplace_title_template: null
> E            md5sum: null
> E            name: fake-image.raw
> E       
> E       
> E       Re-run with PUSHSOURCE_UPDATE_BASELINE=1 if this is correct.
> ```

on it.

JAVGan (approved):
@rohanpm just to let you know: We decided to create an attribute named `marketplace-name` to compose with the `KojiBuildInfo.name` as we thought it wouldn't be good to mess with the `KojiBuildInfo.name` by appending the `-azure` or `-aws` suffix to it, since the information would reflect a build which doesn't exist.

However, we need the build name to have `-azure` or `-aws` for StArMap as the RHCOS names doesn't quite follow the Stratosphere convention names, so instead we're adding these suffix to the new property and resolving it in our `pubtools`.

rohanpm (approved):

PR #476, opened 2024-05-07
Load product metadata from ProductID certificates [RHELDST-24276]

MichalHaluza:
python-rhsm cannot be installed with GCC 14. Given that project has been deprecated for years, it's unlikely the actual underlying issue will be resolved. There are several projects in the distribution realm that used python-rhsm to read ProductID certificates and pull product metadata from them. After some internal discussion, pushsource seems to be an acceptable place where to put the replacement code which can be used instead of python-rhsm.

This commit adds logic for reading metadata from Red Hat ProductID certificate files leveraging `cryptography` and `pyasn1` modules.

It also slightly adjusts `conv.sloppylist()` so the individual string elements are stripped of leading and trailing whitespaces after splitting.


tests/baseline/cases/direct-productid.yml
```diff
@@ -4,7 +4,7 @@
 # To regenerate, run test_baseline.py with PUSHSOURCE_UPDATE_BASELINE=1.
 
 # URL of Source to test.
-url: "productid:/dir/greatest-productid-of-all-time?dest=repo1,repo2,repo3"
+url: "productid:{{ src_dir }}/tests/staged/data/simple_productid/dest1/PRODUCTID/some-cert?dest=repo1,repo2,repo3"
```
MichalHaluza, line 7: @rohanpm: This change was done to prevent the test failing on `FileNotFoundError` as the new code tries to read any ProductID that has its `src` property set. However it leads me to a question - Is it safe to assume that the ProductID files are available and readable at the time when the PushItem obejcts are being generated?

MichalHaluza:
@rohanpm, @rbikar, @amcmahon-rh please review

src/pushsource/_impl/model/productid.py
```diff
@@ -1,14 +1,112 @@
...
+    """
+
+    name = attr.ib(type=str, default=None)
+    """Human readable product name.
+
+    :type: str
+    """
+
+    version = attr.ib(type=str, default=None)
+    """Human readable product version string.
```
rohanpm, line 34: Minor: it would be nice if the doc string had an example value, to give the reader an idea of the expected format in typical scenarios.

src/pushsource/_impl/model/productid.py
```diff
@@ -1,14 +1,112 @@
...
+                #  2 = Product Version, e.g. "9.4"
+                #  3 = Product Architecture, e.g. "s390x"
+                #  4 = Product Tags / Provides, e.g. "rhel-9,rhel-9-s390x"
+                eng_id, attribute_id = map(int, oid.split(".")[9:11])
+                products_data[eng_id][attribute_id] = str(
+                    decoder.decode(extension.value.value)[0]
+                )
+
+        if not products_data:
+            raise ValueError("File is not a ProductID certificate.")
```
rohanpm, line 100: The message should perhaps include `path` because I'm not sure anything else in the backtrace is going to tell us which file failed to load.

src/pushsource/_impl/model/productid.py
```diff
@@ -1,14 +1,112 @@
...
+    architecture = attr.ib(type=list, default=None, converter=convert_maybe(sloppylist))
+    """List of architectures supported by the product.
+
+    :type: List[str]
+    """
+
+    provided_tags = attr.ib(
+        type=list, default=None, converter=convert_maybe(sloppylist)
+    )
+    """List of tags describing the provided platforms used for pairing with other products.
```
rohanpm, line 48: An example here would be nice too.

tests/baseline/cases/direct-productid.yml
rohanpm: While theoretically backwards-incompatible with some scripts we don't know about, I think it is acceptable with reasoning:

- 'staged' source is the only backend used in practice which produces ProductIdPushItem
- when using that source, you definitely have the files available because you discovered the item by crawling the filesystem in the first place.

src/pushsource/_impl/model/productid.py
```diff
@@ -1,14 +1,112 @@
...
 from .. import compat_attr as attr
 
 
+# Red Hat OID namespace is "1.3.6.1.4.1.2312.9",
+# the trailing ".1" designates a Product Certificate.
+OID_NAMESPACE = "1.3.6.1.4.1.2312.9.1."
+
+
+@attr.s()
+class Product(object):
```
rohanpm, line 18: This needs to be exported from `src/pushsource/__init__.py` if it's intended as public API.

src/pushsource/_impl/model/productid.py
```diff
@@ -1,14 +1,114 @@
...
 @attr.s()
 class ProductIdPushItem(PushItem):
     """A :class:`~pushsource.PushItem` representing a product ID certificate.
 
     For push items of this type, the :meth:`~pushsource.PushItem.src` attribute
     refers to a file containing a PEM certificate identifying a product.
+    """
+
+    products = attr.ib(type=list, converter=frozenlist)
+    """List of products described by the ProductID certificate.
```
rohanpm, line 65: Minor: it would be nice if the doc string included:

```
.. versionadded:: 2.45.0
```

...though we have been rather slack with adding these in general.

rohanpm (approved):
The codecov failure would be fixed by https://github.com/release-engineering/pushsource/pull/477.

rbikar (approved):

PR #479, opened 2024-05-11
Retry brew requests [CLOUDDST-22680]

emilyzheng:
non-logged-in brew requests are not retried by default. Enable the retry to reduce push failures caused by brew connection issues.

### DIFF
--- a/src/pushsource/_impl/backend/koji_source.py
+++ b/src/pushsource/_impl/backend/koji_source.py
@@ -248,7 +248,7 @@ class KojiSource(Source):
             tls = self._cache.setdefault("tls", threading.local())
             if not hasattr(tls, "koji_session"):
                 LOG.debug("Creating koji session: %s", self._url)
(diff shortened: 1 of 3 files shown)


src/pushsource/_impl/backend/koji_source.py
```diff
@@ -134,6 +134,7 @@ def __init__(
         timeout=60 * 30,
         cache=None,
         executor=None,
+        koji_opts=None,
```
rohanpm, line 137: Can we please not add this to the public API?  I think it would be fine to simply hardcode that `{"anon_retry": True}` is always used, and not allow the caller to set the options.

Nothing in the API/docs currently promises that `KojiSource` works by using a `koji.ClientSession`, it's only an implementation detail.  Adding this API is exposing that implementation detail and making it potentially harder to change in the future.

rohanpm (changes requested):

rohanpm (approved):

PR #496, opened 2024-05-29
doc: indicate that the list of staged files is an array

ralphbean:
It's an array, right?


rbikar (approved):
yes, it's an array


ref:
* https://github.com/release-engineering/pushsource/blob/db3f53b64679b1c4efcdadc8db3312063a8edb72/src/pushsource/_impl/schema/staged-schema.yaml#L34
* https://github.com/release-engineering/pushsource/blob/db3f53b64679b1c4efcdadc8db3312063a8edb72/src/pushsource/_impl/schema/staged-schema.yaml#L216

PR #558, opened 2024-08-07
Adds filename to pub source. [SPSTRAT-333]

jajreidy:
This adds filename to pub source. To allow an alternate filename for Pub Source. It still returns an AMI push item. There was some slight adjustment needed to AmiPushItem read Push items from differing sources.

### DIFF
--- a/src/pushsource/_impl/backend/pub_source/pub_source.py
+++ b/src/pushsource/_impl/backend/pub_source/pub_source.py
@@ -25,6 +25,7 @@ class PubSource(Source):
         task_id,
         threads=4,
         timeout=60 * 60 * 4,
+        filename="images.json",
     ):
         """Create a new source.
 
@@ -44,6 +45,10 @@ class PubSource(Source):
             timeout (int)
                 Number of seconds after which an error is raised, if no progress is
                 made during queries to Pub.
+
+            filename (str)
+                The name of the file that will be pulled from a pubtask. The default
+                is set to `images.json`.
         """
 
         self._url = force_https(url)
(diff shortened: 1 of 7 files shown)


jajreidy:

src/pushsource/_impl/backend/pub_source/pub_source.py
```diff
@@ -25,6 +25,7 @@ def __init__(
         task_id,
         threads=4,
         timeout=60 * 60 * 4,
+        filename="images.json",
```
rohanpm, line 28: I'm fundamentally disagreeing with this change.

Prior to the change I would describe PubSource as:

- It gives you all the push items found on a particular Pub task.
- The current implementation _happens to_ only give you AMI push items, but that's not something built-in to the API.  It's just that the other push item types which could potentially be supported have not yet been implemented.  They might become supported later, with no API changes needed.

Now after this change it seems to be rather:

- It gives you all the push items _relating to one specified attachment_ on a particular Pub task.

I don't think that is a good API.  It's not going to work for other types of push items which ideally would be supported and don't appear in attached JSON files like AMI do.  It's inconsistent with other sources (e.g. errata source loads all push items on the advisory - it doesn't require the caller to specify some subset of items to be loaded).  The worst is that it forces the caller to know about Pub task implementation details (caller has to know which filename makes sense for their case) rather than keeping it hidden inside this library.

Could you explain the bigger picture of what you want to accomplish?  I read the JIRA issue but it wasn't clear.

If the issue is that some tasks have (only) images.json while others have (only) clouds.json, then I suggest PubSource should try to load both, and use the one which exists (or yield no items if neither exists).  That would be more consistent with the way the other sources work.

rohanpm (changes requested):

src/pushsource/_impl/backend/pub_source/pub_source.py
jajreidy: Thanks for the feedback. I'll close this and reopen one that is more in line with this.

PR #560, opened 2024-08-08
Adds support for clouds.json to Pub Source [SPSTRAT-333]

jajreidy:
Adds support for clouds.json to Pub source. 

The new implementation will try pulling images.json from a Pub task. But if not found then it will try to pull clouds.json from a Pub task.

### DIFF
--- a/src/pushsource/_impl/backend/pub_source/pub_client.py
+++ b/src/pushsource/_impl/backend/pub_source/pub_client.py
@@ -43,19 +43,25 @@ class PubClient(object):
             self._tls.session = requests.Session()
         return self._tls.session
 
-    def _do_request(self, **kwargs):
-        return self._session.request(**kwargs)
+    def _do_request(self, url, **kwargs):
+        kwargs["url"] = url.format(filename="images.json")
+        resp = self._session.request(**kwargs)
+        # if 404 lets try clouds.json
+        if resp.status_code == 404:
+            kwargs["url"] = url.format(filename="clouds.json")
+            resp = self._session.request(**kwargs)
+        return resp
 
     def get_ami_json_f(self, task_id):
         """
         Returns Future[dict|list] holding json obj with AMI push items returned from Pub for given task id.
         """
         endpoint = "pub/task"
-        url_ending = "log/images.json"
+        url_ending = "log/{filename}"
         params = {"format": "raw"}
         url = os.path.join(self._url, endpoint, str(task_id), url_ending)
 
-        LOG.info("Requesting Pub service for %s of task: %s", url_ending, str(task_id))
+        LOG.info("Fetching AMI details from Pub task: %s", str(task_id))
         ft = self._executor.submit(
             self._do_request, method="GET", url=url, params=params
         )
--- a/src/pushsource/_impl/model/ami.py
+++ b/src/pushsource/_impl/model/ami.py
@@ -224,8 +224,8 @@ class AmiPushItem(VMIPushItem):
             "origin": data.get("origin") or None,
             # ami push item fields
             "release": AmiRelease._from_data(data.get("release") or {}),
-            "type": data["type"],
-            "region": data["region"],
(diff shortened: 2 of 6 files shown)


src/pushsource/_impl/backend/pub_source/pub_client.py
```diff
@@ -43,19 +43,25 @@ def _session(self):
...
         Returns Future[dict|list] holding json obj with AMI push items returned from Pub for given task id.
         """
         endpoint = "pub/task"
-        url_ending = "log/images.json"
+        url_ending = "log/{filename}"
         params = {"format": "raw"}
         url = os.path.join(self._url, endpoint, str(task_id), url_ending)
 
-        LOG.info("Requesting Pub service for %s of task: %s", url_ending, str(task_id))
+        LOG.info("Requesting Pub service for JSON file from task: %s", str(task_id))
```
rohanpm, line 64: If you're changing this anyway, might I suggest a log message like "Fetching AMI details from Pub task: %s" makes more sense.

Pub tasks can have JSON files attached for other reasons, so a log message of "Requesting Pub service for JSON file" is a bit vague.

src/pushsource/_impl/model/ami.py
```diff
@@ -236,8 +236,10 @@ def _from_data(cls, data):
             "boot_mode": (
                 BootMode(data.get("boot_mode")) if data.get("boot_mode") else None
             ),
-            "billing_codes": AmiBillingCodes._from_data(
-                data.get("billing_codes") or {}
+            "billing_codes": (
```
rohanpm, line 239: I'm not sure about this particular change here.

The code already had handling for the data being missing.  It would give you a default-constructed `AmiBillingCodes`, which means an AmiBillingCodes object having `name=None` and `codes=[]`.

Is there a reason to change it to instead start generating `None`?  That'd be backwards-incompatible to any code expecting the former behavior.

rohanpm (changes requested):

src/pushsource/_impl/model/ami.py
jajreidy: For some reason if this is missing in the data it throws an error. It seems that optional only wants either an AmiBillingCodes or None. I could change it so that the parameters of AmiBillingCodes are optional and that would probably pass. Is there a better way to do this?

The specific error is name is missing when it goes to parse a missing AmiBillingCodes.

src/pushsource/_impl/backend/pub_source/pub_client.py
jajreidy: That makes sense.

src/pushsource/_impl/model/ami.py
rohanpm: OK, I see.  So the case where it called `AmiBillingCodes._from_data({})` never actually worked.
It surely wasn't intended to be that way, but the good news is it means there's no backwards-compatibility concern with this change.

tests/pub/test_pub_amis.py
```diff
@@ -96,6 +96,189 @@ def test_get_ami_push_items_single_task(requests_mock):
...
+    request_url = os.path.join(
+        pub_url, "pub/task", str(task_id), "log/images.json?format=raw"
+    )
+    requests_mock.register_uri("GET", request_url, status_code=404)
+
+    # request push items from source
+    with Source.get("pub:%s" % pub_url, task_id=task_id) as source:
+        push_items = [item for item in source]
+
+    # there should be exactly one  push item
```
rohanpm, line 187: Comment says one but next line says three.  Could the false comment be fixed please?

Note, if you want to make the comment more "copy-paste resistant" you could word it like "there should be this many push items".  Or even just drop the comment and comparison since the following assertion already implicitly tests the number of push items.

jajreidy: All fixed up

rohanpm (approved):

PR #582, opened 2024-09-03
Fixed some args for AmiPushItem that were marked optional but weren't

jajreidy:
Some of the args in AmiPushItem were set as validating as optional_str but weren't using get method. So if the value didn't exist in whatever it was trying to do .from_data then it would error.


jajreidy:
@JAVGan FYI

jajreidy:

JAVGan (approved):

rohanpm (approved):

rbikar (approved):

PR #627, opened 2024-10-23
Adds staged file structure for cloud images.

jajreidy:
Adds staging for cloud images. The general format for the CLOUD_IMAGES is as follows

* `name of folder` (this is arbitrarily assigned and could be anything)
    * CLOUD_IMAGES
        * resources.yaml
        * image.raw  
                                    
                                    
The resources.yaml contains information that will be used to upload/publish the image. Simple example would be:

```yaml
api: v1
resource: CloudImage
images:
   - path: something1.raw.xz
     architecture: x86_64
   - path: something2.raw.xz
     architecture: arm64
build:
  name: rhel-ec2
  version: "9.4"
  respin: "1"
description: "build1 sample"
boot_mode: hybrid
type: AMI
```

related to: SPSTRAT-421


jajreidy:
@JAVGan FYI

jajreidy:

docs/sources/staged.rst
```diff
@@ -196,6 +197,19 @@ Files in this directory must have metadata included in ``staged.yaml``.
...
 
+root/destination/CLOUD_IMAGES/\*
+..............................
+
+The ``CLOUD_IMAGES`` should contain the VMI(s) plus a ``resources.yaml``.
+
+The ``resources.yaml`` contains all the information needed for the
+images in that folder.
+
+Files in this directory should not include the ``staged.yaml``.
```
rohanpm, line 208:
```suggestion
Files in this directory should not be included within ``staged.yaml``.
```

docs/sources/staged.rst
```diff
@@ -196,6 +197,19 @@ Files in this directory must have metadata included in ``staged.yaml``.
 
 Will yield instances of :class:`~pushsource.AmiPushItem`.
 
+root/destination/CLOUD_IMAGES/\*
+..............................
+
+The ``CLOUD_IMAGES`` should contain the VMI(s) plus a ``resources.yaml``.
```
rohanpm, line 203:
```suggestion
The ``CLOUD_IMAGES`` directory should contain the VMI(s) plus a ``resources.yaml``.
```

src/pushsource/_impl/backend/staged/staged_cloud.py
```diff
@@ -0,0 +1,128 @@
...
+            "src": src,
+            "description": resources.get("description"),
+            "build_info": build_info,
+            "origin": origin,
+        }
+        return VHDPushItem(**image_kwargs)
+
+    @handles_type("CLOUD_IMAGES")
+    def __cloud_push_item(self, leafdir, _, entry):
+        with open(entry.path, "rt") as fh:
```
rohanpm, line 104: This does not seem right; how do you know this is only opening the `resources.yaml`?  Isn't this code going to be executed on every file within the `CLOUD_IMAGES` directory?

The tests are probably only able to pass because every sample image you added is an empty file, and `yaml.safe_load` will happily accept an empty string and return `None`.  I think if you added any non-empty VMI to the test data then the tests would crash, detecting the bug here.  So can you please do that?  (doesn't have to be a real VMI.)

src/pushsource/_impl/backend/staged/staged_cloud.py
```diff
@@ -0,0 +1,128 @@
...
+            raw = yaml.safe_load(fh)
+        if not raw:
+            return
+
+        out = []
+        image_type = raw.get("type")
+        images_info = raw.get("images")
+
+        for image in images_info:
+            image_relative_path = os.path.join(leafdir.path, image.get("path"))
```
rohanpm, line 114: I would like it to be clarified whether `path` is allowed to point outside of the current directory.
e.g. would it be allowed for it to be "../../other-dir/my-image.xz".

I'd rather we don't allow this (unless you really know that something will need it).  i.e., please enforce that `path` does not contain a `/` character.

If it later turns out that you need it, we can always loosen this restriction.  But if we allow it from the beginning, adding the restriction later is potentially backwards-incompatible.

docs/sources/staged.rst
```diff
@@ -196,6 +197,19 @@ Files in this directory must have metadata included in ``staged.yaml``.
...
+
+The ``resources.yaml`` contains all the information needed for the
```
rohanpm, line 205: The format of this will have to be documented somewhere.
Ideally it should have a schema, like the existing `errata-schema.yaml` and `staged-schema.yaml`, which could be linked from here.

docs/sources/staged.rst
```diff
@@ -196,6 +197,19 @@ Files in this directory must have metadata included in ``staged.yaml``.
 
 Will yield instances of :class:`~pushsource.AmiPushItem`.
 
+root/destination/CLOUD_IMAGES/\*
```
rohanpm, line 200: What's implemented doesn't match what's in the design doc.

Design doc: https://url.corp.redhat.com/ab

The idea was there would be dirs like `root/<destination>/CLOUD_IMAGES/<build>`, where each "build" dir contains a "resources.yaml" and referenced images. 

In practice, there would always only be one destination - starmap - while there would be multiple builds.

For instance the doc had: "Example of a valid staged directory with AmiPushItems for RHEL 9.0, 9.4 and 9.6":

```
/starmap/CLOUD_IMAGES/
  /rhel-ec2-9.0-20241010.1/
    resources.yaml
    rhel-ec2-9.0-20241010.1.x86_64.raw.xz
  /rhel-ec2-9.4-20241010.1/
    resources.yaml
    rhel-ec2-9.4-20241010.1.x86_64.raw.xz
  /rhel-ec2-9.6-20241010.1/
    resources.yaml
    rhel-ec2-9.6-20241010.1.x86_64.raw.xz
```

What you've implemented instead seems to be this:

- dirs `root/<destination>/CLOUD_IMAGES/` exist
- the CLOUD_IMAGES directory itself contains `resources.yaml` and the images (no further subdir)
- In order to include multiple `resources.yaml`, you can have multiple different destinations.  For instance in your test data you used `build1`, `build2`, `build3` as some destinations.

Is it intentionally different from the design?  I think the agreed-upon design in the doc was better.

rohanpm (changes requested):

docs/sources/staged.rst
jajreidy: I've reworked this to match the agreed design in the doc. I did have to include an empty staged.yaml file otherwise it would be skipped in the scan.

src/pushsource/_impl/backend/staged/staged_cloud.py
```diff
@@ -0,0 +1,141 @@
...
+                "type",
+            ]
+            release_kwargs = {}
+            for key in release_attrs:
+                release_kwargs[key] = release_resources.get(key)
+
+            image_kwargs["release"] = AmiRelease(**release_kwargs)
+
+        image_attrs = [
+            "dest",
```
rohanpm, line 57: Can we please drop "dest" from this list?
Staged source is defined as deriving the "dest" (destination) from the path.
I would rather this be universally consistent and we don't introduce a different way of setting the dest for this content type.

I believe that your client code is expected to ignore the value anyway and set it from elsewhere after the item comes out of pushsource library, so I think this shouldn't affect you, right?

src/pushsource/_impl/backend/staged/staged_cloud.py
```diff
@@ -0,0 +1,141 @@
...
+        for bd in build_dirs:
+            yaml_path = os.path.join(bd, "resources.yaml")
+            try:
+                with open(yaml_path, "rt") as fh:
+                    raw = yaml.safe_load(fh)
+            except FileNotFoundError:
+                LOG.warning("No resources.yaml file found at %s (ignored)", yaml_path)
+                continue
+            if not raw:
+                continue
```
rohanpm, line 113: Perhaps this would also warrant a warning?  It's just a little odd that a missing file generates a warning but a present-but-empty file doesn't.

src/pushsource/_impl/backend/staged/staged_cloud.py
```diff
@@ -0,0 +1,141 @@
...
+            image_type = raw.get("type", "")
+            images_info = raw.get("images", [])
```
rohanpm, line 115: Up to you, but I would rather encourage this style for defaults:

```suggestion
            images_info = raw.get("images") or []
```

...because the way you have it right now will crash if `images` is present and `null`.

src/pushsource/_impl/backend/staged/staged_cloud.py
```diff
@@ -0,0 +1,141 @@
...
+                    out.append(
+                        self.__build_azure_push_item(
+                            raw, image_relative_path, bd, image.get("path")
+                        )
+                    )
+        return out
+
+    @handles_type("CLOUD_IMAGES")
+    def __cloud_push_item(self, leafdir, _, __):
+        builds = [f.path for f in os.scandir(leafdir.path) if f.is_dir()]
```
rohanpm, line 138: Regarding the "have to provide an empty staged.yaml" issue, here's the problem I think you're having.

The implementation of `@handles_type` expects that every type handler is designed to deal with *files* only. Until your change, that was true everywhere.

Handlers are invoked like this:

```
    def __mixin_push_items(self, leafdir, metadata, delegate):
        out = []

        LOG.debug("Looking for files in %s", leafdir)

        for entry in scandir(leafdir.path):
            if entry.is_file():
                item = delegate(leafdir, metadata, entry)
                if item:
                    out.append(item)

        return out
```

``leafdir`` is ``<root>/destination/CLOUD_IMAGES``.  Also note that `scandir` isn't recursive.  So, it's designed to call the handler once for every file inside the leafdir (non-recursive).  In your case, there are zero files within the leafdir - only directories - so it doesn't do anything.  You can force it to do something by adding an empty `staged.yaml` - or adding *any* file in the directory for that matter (and one wonders what it's going to do if the user adds more than one file).

Making the user add a file into the directory just to satisfy this code isn't acceptable, the code should rather be modified to better suit your scenario.

I have two suggested approaches how to fix this:

### Suggestion 1

Move the responsibility for the "is_file" checks *inside* of the handlers.

So, the code invoking the handlers becomes like this:

```
        for entry in scandir(leafdir.path):
             item = delegate(leafdir, metadata, entry)
             if item:
                 out.append(item)
```

All existing handlers internally implement "if entry is not a file, do nothing" like this:

```
    @handles_type("COMPS")
    def __push_item(self, leafdir, _, entry):
        if entry.is_file():
          return CompsXmlPushItem(
              name=entry.name, src=entry.path, origin=leafdir.topdir, dest=[leafdir.dest]
          )
```

And your own handler implements "if entry is not a dir containing resources.yaml, do nothing" like this:

```
    @handles_type("CLOUD_IMAGES")
    def __cloud_push_item(self, leafdir, _, entry):
        if not entry.is_dir() or not os.path.exists(os.path.join(entry.path, "resources.yaml"):
            return
        # ...OK, we know `entry` is a dir containing a resources.yaml, process it
```

### Suggestion 2

Update the `@handles_type` decorator to take a parameter controlling the check for whether or not a dir entry should be processed.

Your own handler then might look something like this:

```
    @handles_type("CLOUD_IMAGES",
        accepts=lambda entry: entry.is_dir() and os.path.exists(os.path.join(entry.path, "resources.yaml")))
    def __cloud_push_item(self, leafdir, _, entry):
        # When we get here, we already know `entry` is a dir containing a resources.yaml
```

The default value of `accepts` could be set to `lambda entry: entry.is_file()` so that all existing handlers would continue to work how they do today, without having to touch their code.

tests/baseline/cases/staged-simple-cloud.yml
```diff
@@ -0,0 +1,193 @@
...
+    billing_codes: null
+    boot_mode: hybrid
+    build: null
+    build_info:
+      id: null
+      name: rhel-ec2
+      release: '1'
+      version: '9.4'
+    description: build2 sample
+    dest: []
```
rohanpm, line 22: I don't think these `dest` should be empty. I know in the case of your client you're going to ignore the values anyway and overwrite them, which is fine, but I think the behavior in pushsource library should still be consistent for all content types produced by "staged" source.

i.e., can you please have the implementation fill in this value from the path, the same way it works elsewhere? (Meaning the value here should be `["starmap"]` .)

tests/baseline/cases/staged-simple-cloud.yml
```diff
@@ -0,0 +1,193 @@
...
+    origin: {{ src_dir }}/tests/staged/data/simple_cloud/starmap/CLOUD_IMAGES/build2
+    public_image: null
+    recommended_instance_type: null
+    region: null
+    release: null
+    release_notes: null
+    root_device: null
+    scanning_port: null
+    security_groups: []
+    sha256sum: null
```
rohanpm, line 39: Something I've just noticed: in the resources.yaml there are no checksums.  That's pretty unusual.  For most kinds of push items supported today, there is some form of support for the expected checksums to be passed into pushsource library, allowing the content to be verified while it's loaded.

Maybe that should be the case here as well, with resources.yaml having a sha256sum which is then propagated into the push item.

Can you discuss in your team whether this should be added? I can't say it's strictly required, but if omitted some of the consequences might be:

- in case something unexpected/corrupted gets pushed, it will be ambiguous which component was at fault
- in case you want to reliably test whether something has already been uploaded to a remote service, you will have to read the entire image file to calculate its checksum in order to do that check.  You can't cheaply compare local and remote checksums, as you have no local checksums.

src/pushsource/_impl/backend/staged/staged_cloud.py
```diff
@@ -0,0 +1,141 @@
...
+from .staged_base import StagedBaseMixin, handles_type
+from ...model import VHDPushItem, AmiPushItem, AmiRelease, BootMode, KojiBuildInfo
+
+LOG = logging.getLogger("pushsource")
+
+
+class StagedCloudMixin(StagedBaseMixin):
+    def __build_ami_push_item(self, resources, src, origin, name):
+        build_resources = resources.get("build")
+        release_resources = resources.get("release", {})
```
rohanpm, line 14: As this will crash later if `release` is present but null, I suggest this more robust method of setting the default:

```suggestion
        release_resources = resources.get("release") or {}
```

rohanpm (changes requested):

tests/baseline/cases/staged-simple-cloud.yml
jajreidy: Discussed it with the team and it makes sense. So I added it.

docs/sources/staged.rst
```diff
@@ -196,6 +197,22 @@ Files in this directory must have metadata included in ``staged.yaml``.
...
+
+The ``CLOUD_IMAGES`` directory should contain the VMI(s) plus a ``resources.yaml``.
+
+The ``resources.yaml`` contains all the information needed for the
+images in that folder.
+
+:ref:`cloud_schema` provides a complete reference of the fields which can be set by this
+file.
+
+A blank ``staged.yaml`` must be included in the root directory.
```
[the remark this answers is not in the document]
jajreidy: Removed as it isn't needed (and was a bad hack on my part) in favor of updating the handler.

rohanpm (approved):

src/pushsource/_impl/backend/staged/staged_cloud.py
```diff
@@ -0,0 +1,136 @@
...
+        image_kwargs = {
+            "name": name,
+            "src": src,
+            "description": resources.get("description"),
+            "build_info": build_info,
+            "origin": origin,
+            "dest": [dest],
+            "sha256sum": image.get("sha256sum"),
+        }
+        return VHDPushItem(**image_kwargs)
```
JAVGan, line 107: Sorry I didn't notice this before but it seems you're not creating the `VMIRelease` for `VHDPushItem`, this might break the publishing as you can [see here](https://github.com/release-engineering/pubtools-marketplacesvm/blob/main/src/pubtools/_marketplacesvm/cloud_providers/ms_azure.py#L105-L144) and [here](https://github.com/release-engineering/pubtools-marketplacesvm/blob/main/src/pubtools/_marketplacesvm/cloud_providers/ms_azure.py#L305)

jajreidy: Oh, we will need to add date and product then.

JAVGan: tyvm

rohanpm (approved):

PR #636, opened 2024-11-04
Fixes release to include version from build.

jajreidy:
When adding cloud to staged files version was missed as a requirement for releases. This adds that in to avoid issues with Stratosphere releases.

### DIFF
--- a/src/pushsource/_impl/backend/staged/staged_cloud.py
+++ b/src/pushsource/_impl/backend/staged/staged_cloud.py
@@ -57,12 +57,12 @@ class StagedCloudMixin(StagedBaseMixin):
 
         release_kwargs = {
             "product": self.__get_product_name(build_resources.get("name")),
-            "date": datetime.now(timezone.utc).strftime("%Y%m%d"),
+            "date": release_resources.get("date") or datetime.now(timezone.utc).strftime("%Y%m%d"),
             "arch": image.get("architecture"),
             "respin": int(build_resources.get("respin")) or 0,
+            "version": build_resources.get("version"),
         }
         release_attrs = [
-            "version",
             "base_product",
             "base_version",
             "variant",
@@ -104,6 +104,7 @@ class StagedCloudMixin(StagedBaseMixin):
 
     def __build_azure_push_item(self, resources, origin, image, dest):
         build_resources = resources.get("build")
+        release_resources = resources.get("release") or {}
         name = image.get("path")
         src = os.path.join(origin, name)
         build_info = KojiBuildInfo(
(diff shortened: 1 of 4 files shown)


jajreidy:
@JAVGan FYI

jajreidy:

JAVGan (approved):
LGTM!

Just for the context, we may need the build time to be set in `resources.yaml` as it might be used for composing the image name, which will require the same value for all pushes to be indepontent, preventing the tooling to upload multiple times the same image just because the time did change.

src/pushsource/_impl/backend/staged/staged_cloud.py
```diff
@@ -57,12 +57,14 @@ def __build_ami_push_item(self, resources, origin, image, dest):
 
         release_kwargs = {
             "product": self.__get_product_name(build_resources.get("name")),
-            "date": datetime.now(timezone.utc).strftime("%Y%m%d"),
+            "date": release_resources.get("date")
+            or datetime.now(timezone.utc).strftime("%Y%m%d"),
```
rohanpm, line 61: I noticed the `datetime.now` caused problems for tests after the original change was submitted.

I'm not sure if it should be fixed just for the tests though, maybe it should rather be fixed in general by defaulting this to `None` or making it mandatory?

The problem with the tests was that loading the same data on different days will give different results.  But that seems like a bad thing in general, not just for the tests.  I don't think it's desirable that loading a staging area today might give you different data than loading it tomorrow even when there were no changes.

The default logic also does NOT match what the doc string says for this field, which is: `"Date at which this image was generated"`.  It's not valid to assume that's today when the field is missing.

If there is to be any such logic of "when the date is missing, use today", I'd feel more comfortable if that's happening in the downstream project making use of these push items.  Downstream is in a better position to decide whether applying such a default is reasonable.

rohanpm: I just read @JAVGan's last comment, isn't he saying much the same?  It is not desirable that the "date" field can change depending on which day you load the staging area, as that could mean that (say) an attempted push on one day, retried the next day, could end up pushing two different image names.

rohanpm (changes requested):

src/pushsource/_impl/backend/staged/staged_cloud.py
jajreidy: Makes sense, I removed the date generation from this.

src/pushsource/_impl/backend/staged/staged_cloud.py
```diff
@@ -57,12 +57,13 @@ def __build_ami_push_item(self, resources, origin, image, dest):
 
         release_kwargs = {
             "product": self.__get_product_name(build_resources.get("name")),
-            "date": datetime.now(timezone.utc).strftime("%Y%m%d"),
```
rohanpm, line 60: The now-unused `datetime` import needs to be removed for this to pass pylint.

rohanpm (approved):

PR #639, opened 2024-11-14
Adds vhd files as a valid target for pub source.

jajreidy:
Adds support for VHD sourced files when using pub as a source. 

Related to: SPSTRAT-362

### DIFF
--- a/src/pushsource/_impl/backend/pub_source/pub_client.py
+++ b/src/pushsource/_impl/backend/pub_source/pub_client.py
@@ -54,14 +54,15 @@ class PubClient(object):
 
     def get_json_f(self, task_id):
         """
-        Returns Future[dict|list] holding json obj with AMI push items returned from Pub for given task id.
+        Returns Future[dict|list] holding json obj with AMI or VHD
+        push items returned from Pub for given task id.
         """
         endpoint = "pub/task"
         url_ending = "log/{filename}"
         params = {"format": "raw"}
         url = os.path.join(self._url, endpoint, str(task_id), url_ending)
 
-        LOG.info("Fetching AMI details from Pub task: %s", str(task_id))
+        LOG.info("Fetching AMI/VHD details from Pub task: %s", str(task_id))
         ft = self._executor.submit(
             self._do_request, method="GET", url=url, params=params
         )
--- a/src/pushsource/_impl/backend/pub_source/pub_client.py
+++ b/src/pushsource/_impl/backend/pub_source/pub_client.py
@@ -52,7 +52,7 @@ class PubClient(object):
             resp = self._session.request(**kwargs)
         return resp
 
-    def get_ami_json_f(self, task_id):
+    def get_json_f(self, task_id):
         """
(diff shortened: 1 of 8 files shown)


jajreidy:
@rohanpm PTAL

@JAVGan FYI

src/pushsource/_impl/backend/pub_source/pub_client.py
```diff
@@ -52,7 +52,7 @@ def _do_request(self, url, **kwargs):
             resp = self._session.request(**kwargs)
         return resp
 
-    def get_ami_json_f(self, task_id):
+    def get_json_f(self, task_id):
         """
         Returns Future[dict|list] holding json obj with AMI push items returned from Pub for given task id.
         """
```
JAVGan, line 58: Please update docstrings and the `LOG.info` message as well 😄

JAVGan: You may replace `AMI` push items with `VMI` push items, or VM images...

JAVGan (approved):
LGTM

src/pushsource/_impl/backend/pub_source/pub_source.py
```diff
@@ -73,26 +73,38 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
         out = []
-
-        try:
-            push_items = AmiPushItem._from_data(json_obj)
-            out.extend(list_argument(push_items))
-        except KeyError:
-            LOG.warning("Cannot parse AMI push item/s: %s", str(json_obj))
+        first_obj = next(iter(json_obj or []), None)
+        if first_obj:
+            if first_obj.get("src").endswith("vhd") or first_obj.get("src").endswith(
```
rohanpm, line 80: This somewhat breaks the intended behavior of the later try/excepts of being tolerant against unknown data.

For example, if the objects had no `"src"` at all, I suppose that would have previously fallen into the "Cannot parse" warning case which is non-fatal.  But now due to this expression it would instead crash with: `'NoneType' object has no attribute 'endswith'`.

Given that there is no way to enforce the format of the files being attached to Pub tasks, that the format might change over time, and that the Pub source might eventually be used for non-AMI/VMI cases, I think it'd be best to continue being more tolerant here.  Can you find a way to write this which won't crash when "src" is not a string?

src/pushsource/_impl/model/azure.py
```diff
@@ -65,3 +65,31 @@ def __attrs_post_init__(self):
                 'The attribute "legacy_sku_id" must only be set when'
                 ' "support_legacy" is True.'
             )
+
+    @classmethod
+    def _from_data(cls, data):
+        """Instantiate AmiPushItem from raw list or dict"""
```
rohanpm, line 71: The doc string is false, referring to the wrong class.

It's reasonable to not even bother with a comment / doc string like this, where it's a private method and the purpose of the method is easy to figure out anyway.  A comment which doesn't exist can't be wrong.

rohanpm (changes requested):

rohanpm (approved):

PR #671, opened 2025-08-26
VMIPushItem: Use converter for boot_mode

JAVGan:
This commit changes the `VMIPushItem` class to use a converter in the attribute `boot_mode` which is capable of properly converting it to the `BootMode` enum whenever possible.

It fixes issues downstream where the information is given as a string instead of the Enum.

Refers to SPSTRAT-600


JAVGan:
@lslebodn @rbikar PTAL

rbikar (approved):

JAVGan:
@rbikar feel free to merge it as I don't have permission to do it

PR #676, opened 2025-09-19
bugfix(SPSTRAT-601): VHDPushItem model _from_data

JAVGan:
This commit fixes a bad converter for `support_legacy` on `VHDPushItem` model which was preventing to load its class using `_from_data` whenever the value for `support_legacy` was `False`.

In the previous version the following line would evaluate the truthful to `else` and cause the `or` to define it as `None`, generating an invalid parameter:

```
"support_legacy": data.get("support_legacy") or None,
```

Besides the fix it also includes a new model test to detect this scenario as well as a test on `Pub` source which reproduces this issue in the previous code.

Refers to SPSTRAT-601


JAVGan:
@rbikar feel free to merge it as I don't have permissions 🙂

PR #714, opened 2026-06-10
feat: add a new source type named "coreos"

JAVGan:
This commit adds a new source type named "coreos" which is capable of fetching push items from the openshift installer JSON in GitHub.

With this, it's possible to yield `AMIPushItem` and `VHDPushItem` from the installer, as shown in the example below:

```
pushsource-ls "coreos:https://github.com/openshift/installer/tree/release-4.22?paths=data/data/coreos/coreos-rhel-10.json"
```

The new entrypoint also support multiple paths, as shown below:

```
pushsource-ls "coreos:https://github.com/openshift/installer/tree/release-4.22?paths=data/data/coreos/coreos-rhel-9.json,data/data/coreos/coreos-rhel-10.json"
```

This is part of an effort to unblock the releases of new OpenShift versions (greater than v4.19)

Signed-off-by: Jonathan Gangi <jgangi@redhat.com>

Assisted-by: Cursor

### DIFF
--- a/docs/sources/coreos.rst
+++ b/docs/sources/coreos.rst
@@ -1,0 +1,31 @@
+Source: coreos
+==============
+
+The ``coreos`` push source allows the loading of content from a JSON file of OpenShift CoreOS installer from GitHub.
+
+Supported content types:
+
+* AMIs
+* VHDs
+
+
+coreos source URLs
+------------------
+
+The base form of an coreos source URL is:
+
+``coreos:github-url?paths=path1[,path2[,...]]``
+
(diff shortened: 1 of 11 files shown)


JAVGan:
@ashwgit @lslebodn can you PTAL?

I'm planning on adding this to unblock the RHCOS releases, at least semi-manually.

With that we may be able to invoke `pubtools-marketplacesvm` with this new source.

NOTE: This is also related to https://github.com/release-engineering/pubtools-marketplacesvm/pull/150

CC: @rbikar

PR #716, opened 2026-06-24
Add Pulp 3 RPM resolution to KonfluxSource [RHELDST-38332]

crungehottman:
KonfluxSource now queries Pulp 3 by SHA256 to resolve each RPM's pulp_href, which is set as RpmPushItem.src for Pulp-to-Pulp syncing.

Changes:
- Add Pulp3Client integration: look up RPMs by SHA256 checksum and set src=pulp_href on RpmPushItem
- Accept Pulp connection params (pulp_url, pulp_cert, pulp_key, pulp_domain) as constructor args via Source.get() query params
- Auto-discover advisory subdirectories instead of requiring an explicit advisories parameter
- Add pubtools-pulplib>=2.44.0 as a required dependency

### DIFF
--- a/src/pushsource/_impl/backend/konflux_source/konflux_source.py
+++ b/src/pushsource/_impl/backend/konflux_source/konflux_source.py
@@ -1,5 +1,7 @@
 import logging
+import os
 
+import anyio
 from more_executors import Executors
 
 from ...source import Source
@@ -8,6 +10,8 @@ from ... import compat_attr as attr
 from ...helpers import list_argument, as_completed_with_timeout_reset
 
 from .konflux_loader import KonfluxLoader
+
+from pubtools.pulplib import Pulp3Client
 
 LOG = logging.getLogger("pushsource.konflux")
 
@@ -30,6 +34,12 @@ class KonfluxSource(Source):
         self,
         url,
         advisories,
+        pulp_url=None,
+        pulp_cert=None,
+        pulp_key=None,
+        pulp_user=None,
+        pulp_password=None,
+        pulp_domain=None,
         threads=4,
         timeout=60 * 60,
     ):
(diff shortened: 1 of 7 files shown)


src/pushsource/_impl/backend/konflux_source/konflux_source.py
```diff
@@ -9,6 +11,8 @@
 
 from .konflux_loader import KonfluxLoader
 
+from pubtools.pulplib._impl.client_pulp3 import Pulp3Client
```
rbikar: we should be able to import this also as
`from pubtools.pulplib import Pulp3Client`

src/pushsource/_impl/backend/konflux_source/konflux_source.py
```diff
@@ -30,6 +34,10 @@ def __init__(
         self,
         url,
         advisories,
+        pulp_url=None,
```
rbikar, line 37: I believe the recommended way of accessing hosted-pulp is via user or service account using standard username:password. So I believe we should also support username:password auth.

src/pushsource/_impl/backend/konflux_source/konflux_source.py
```diff
@@ -180,11 +244,66 @@ def _create_rpm_items(self, data):
 
         return items
 
+    async def _query_rpm_in_pulp(self, sha256sum):
```
rbikar: this doesn't look super effective, this instantiates `Pulp3Client` for each rpm that's required and queries RPMs one by one.

1) Could we make the pulp3 client embedded in this class in a similar way as e.g. errata source? At least use only one instance of client for whole `_create_rpm_items()`
2) I'd rather try and implement querying for multiple RPMs at once + support some batches. 

But I guess the current solution would be fine as some starting point and we can do improvements later.

rbikar (changes requested):

src/pushsource/_impl/backend/konflux_source/konflux_source.py
crungehottman: Agreed. I wrote a comment in the Jira with more details, but it seems like the maximum batch size is 7 (for now, at least), as a query with 8 checksums fails with `Filter expression exceeds allowed complexity`.

rbikar (approved):

